PR #4, opened 2019-01-12
Introduce tests & tooling

antonve:
## Why

Testing is essential for a high quality code base.

## What

- [x] Use [gex](https://github.com/izumin5210/gex) to manage dependencies that are only needed to run dev tools, like [mockgen](https://github.com/golang/mock/)
- [x] Introduce [mockgen](https://github.com/golang/mock/)
- [x] Add a `Makefile` to document the most used commands to work with this codebase
- [x] Add sample spec for the `HealthService`
- [x] Update README
- [x] Enable go modules by default


antonve:
cc @Hobogrammer

PR #12, opened 2019-01-20
Registration endpoint

antonve:
### Architectural
- [x] Setup repository & interactor pattern
- [x] Clean up DI for repositories, interactors, and services
- [x] Fix row/rows distinction in `SqlHandler`

### Testing
- [x] Setup integration tests with actual database for repositories
- [x] Add testing configuration
- [x] Add spec for `UserRepository`

### Etc
- [x] Implement `UserRepository`
- [x] Implement `SessionInteractor`
- [x] Add scanning/value interfaces for `Preferences`: I'm not too happy that there is a reference to `database/sql` in here, but it doesn't seem like we can avoid this as we can't add methods to structs from a different package


antonve:
cc @Hobogrammer

PR #13, opened 2019-01-21
Add password hashing

antonve:
## Why

Plaintext password storing shouldn't be kept due that being a *very bad idea (tm)*.

## What

- [x] Implement bcrypt hashing
- [x] Update tests with appropriate mock


antonve:
cc @Hobogrammer

PR #14, opened 2019-01-21
Login endpoint

antonve:
## What

I've opted to use JWT tokens as that seems to be quite standard.

- [x] Move password hasher from `UserRepository` to `SessionInteractor` as it's supposed to be part of the business logic, not database level
- [x] Implement JWT token generator 
- [x] Check for errors when configuring the server, it will now panic if not all configurations are set
- [x] Prevent password from being exposed in JSON without having to remember it all over the place
- [x] Add compare functionality to `PasswordHasher` so we can check if it's the same password
- [x] Implement `UserRepository.FindByEmail`, this will return the password as it's needed to check equality during login
- [x] Implement `Login` endpoint in the `SessionsService`

## Example

```sh
$ curl --request POST \
  --url http://localhost:3000/login \
  --header 'content-type: application/json' \
  --data '{
	"email": "test@test.com",
	"password": "testtest"
}'
{
  "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1NDgxNjQxNTUsInVzZXIiOnsiaWQiOjgsImVtYWlsIjoidGVzdEB0ZXN0LmNvbSIsImRpc3BsYXlfbmFtZSI6InRlc3QiLCJwYXNzd29yZCI6IiIsInJvbGUiOjEsInByZWZlcmVuY2VzIjp7fX19.XxgSS1jb8h0Ajn73JA3Y69JB1H4z3UQuivS3cELOu_o",
  "user": {
    "id": 8,
    "email": "test@test.com",
    "display_name": "test",
    "password": "",
    "role": 1,
    "preferences": {}
  }
}
```
## Caveats

No tests for the interactors or session service have been written so far, will do in a follow-up PR.
Ideally these tokens should be short-lived and be refreshable with a long-living token that can be blacklisted. At the moment I've not implemented this yet. This is something we should take care of before we go live, but the current implementation lets us move forward.


antonve:
cc @Hobogrammer

PR #29, opened 2019-01-28
Add JWT authentication to router

antonve:
## Why

To close https://github.com/tadoku/api/issues/17

## What

- [x] Research JWT in echo: https://echo.labstack.com/cookbook/jwt
- [x] Figure out how to abstract that away with the current implementation of routes
- [x] Implement a solution so it can be easily reused
- [x] Implement spec

### DIFF
--- a/infra/router_test.go
+++ b/infra/router_test.go
@@ -1,0 +1,44 @@
+package infra_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/tadoku/api/infra"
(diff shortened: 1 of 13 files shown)


infra/router_test.go
```diff
@@ -0,0 +1,44 @@
...
+	e := infra.NewRouter("1337", secret, routes...)
+
+	for _, tc := range []struct {
+		path          string
+		expStatusCode int
+	}{
+		{
+			path: "/restricted",
+			// @TODO: This should be 401 instead of 400
+			expStatusCode: http.StatusBadRequest,
```
antonve: 📝 https://github.com/tadoku/api/issues/30

PR #47, opened 2019-05-01
Update rankings when logging pages

antonve:


usecases/repositories.go
```diff
@@ -27,5 +27,7 @@ type ContestLogRepository interface {
 // RankingRepository handles Ranking related database interactions
 type RankingRepository interface {
 	Store(contest domain.Ranking) error
+	FindAll(contestID uint64, userID uint64) (domain.Rankings, error)
 	GetAllLanguagesForContestAndUser(contestID uint64, userID uint64) (domain.LanguageCodes, error)
+	UpdateRankingsForContestAndUser(contestID uint64, userID uint64) error
```
antonve: This should live in `RankingInteractor`.

In the repository we need an `UpdateRankings` which takes in an array of rankings, to be updated in a transaction.

PR #315, opened 2021-11-14
Show footer social links on mobile

sheodox:
On screen sizes under a mobile breakpoint the social (twitter/github/discord) links were hidden. I replaced that styling with some styles to center the icons to match the centered Tadoku logo in the footer.

The "Built by antonve" link was given a similar treatment, but the only change was to make the margins to be symmetrical instead of the bottom margin being twice the top. This looks better in my opinion when everything is closer. The wider breakpoint styles have 0px margins on the left/right sides, but I didn't bother with that here, the text won't wrap unless the viewport is already unreasonably narrow.
                                                                                              
image


antonve (approved):
Thanks!

PR #386, opened 2022-02-28
Fix user profile update and password changing

sheodox:
The default api endpoint exported from `api.ts`'s `post`/`get`/`destroy`/`put` exports is for interacting the contest. The user settings page imports the `post` and tries to use that for user changes, but it should be using the identity service instead. So currently when you try to update your username it posts to `/api/reading-contest/users/2/profile` which obviously 404s.

This change just switches to using the identity api client instead.


antonve (approved):
Thanks!

PR #413, opened 2022-04-19
Added a modal to generate contest stats for a blog post

sheodox:
This adds a modal with a textarea that shows some blog post html for a contest with all the generic text that'd be needed.

The modal just does a bunch of stuff with document.createElement() then returns the innerHTML of a temporary div so we should avoid XSS if it somehow managed to get in there (I haven't checked the username validation rules, but I imagine with Japanese names I've seen it is probably fairly permissive).

This required a new endpoint for admin roles, and I used the queries you sent me but the response is still kind of messy and doesn't conform to the snake case style used by other http routes. This is still pretty WIP so I don't expect this to merge as-is.

I added an editorconfig file so nvim stops using tabs for TS files. I also added a package.json script that runs prettier because until I ran that I had a lot of code that didn't conform to prettier's style.


frontend/web/package.json
```diff
@@ -13,6 +13,7 @@
     "test": "jest",
     "lint": "eslint --ext .js,.ts,.tsx ./",
     "lint-fix": "eslint --fix --ext .js,.ts,.tsx ./",
+    "style-fix": "prettier --write .",
```
antonve, line 16: Suggestion: Can we name this `fmt`? That way all formatting in this project use the same command (like `go fmt`.

frontend/web/src/app/contest/components/modals/ViewStatsModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+          el.setAttribute(attr, attributes[attr])
+        }
+      }
+    }
+    return el
+  }
+
+  const winner = ranking[0]
+
+  return el('div', '', [
```
antonve, line 36: This I would prefer to be written in idiomatic React code. Seems like it's re-implementing what React already does for us? `renderToString(element)` can be used to get the raw html from a React component.

Ref. https://reactjs.org/docs/react-dom-server.html#rendertostring

services/reading-contest-api/domain/contest.go
```diff
@@ -43,3 +43,14 @@ type ContestID uint64
 func (id ContestID) IsGlobal() bool {
 	return id == 0
 }
+
+type ContestStats struct {
+	ByLanguage   []ContestLanguageStat `json:"byLanguage"`
+	Participants int                   `json:"participants"`
+	TotalPages   float64               `json:"totalPages"`
```
antonve, line 50: Can we call this `TotalAmount`? `Amount` is what we've been using in the code base to refer to a user's score/pages.

services/reading-contest-api/interfaces/repositories/contest_repository.go
```diff
@@ -138,3 +138,62 @@ func (r *contestRepository) FindByID(id uint64) (domain.Contest, error) {
 
 	return contest, nil
 }
+
+func (r *contestRepository) Stats(id uint64) (domain.ContestStats, error) {
+	var contestStats domain.ContestStats
```
antonve, line 143: For other features i'd most likely split out individual queries into different functions, but it's ok for this one. I don't see this one being changed all that often in the near future.

frontend/web/src/app/contest/components/modals/ViewStatsModal.tsx
sheodox: Done! Though I used `renderToStaticMarkup` which is the same thing but without any react attributes. https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup

antonve (approved):
Thanks!

PR #421, opened 2022-04-21
Fix the leaderboard being wider than the screen when loading

sheodox:
When the ranking is loading and skeletons are shown the rankings were way too wide, so wide the ranking would go off the right edge of the page and the recent updates column would be wrapped below. It seems that removing `flex-basis: auto` from `Content` fixes it. I believe the 100% width on on the Skeleton in Leaderboard.tsx is what was causing issues with the flex basis.

The bug:
image

The fix:
image

### DIFF
--- a/frontend/web/src/app/ranking/pages/RankingOverview.tsx
+++ b/frontend/web/src/app/ranking/pages/RankingOverview.tsx
@@ -179,7 +179,7 @@ const TwoColumn = styled.div`
 `
 
 const Content = styled.div`
-  flex: 1 0 auto; // enable grow, disable shrink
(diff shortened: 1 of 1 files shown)


sheodox:
https://github.com/tadoku/tadoku/blob/f5fb8b39e3b37b75d9d3868be91486ac00496587/frontend/web/src/app/ranking/components/Leaderboard.tsx#L222-L225

Those lines are also causing the nickname skeleton to not be vertically aligned with the other skeletons. Not sure if I should address that in this or not.

antonve (approved):

PR #622, opened 2025-04-16
fixed error in manual.tsx

JaiWWW:
the stated multiplier (0.0008333...) which is also the modifier actually used in character to page conversion represents 1200 characters per page, not 1800


antonve (approved):
Thanks!

PR #704, opened 2026-07-24
Display duration across log views

antonve:
## Summary

- add shared formatting for amount-only, duration-only, and combined tracking values
- display time spent in log lists, admin and non-admin detail views, and contest submission
- include `duration_seconds` in contest and profile log-list API responses
- add focused regression coverage for both list endpoints
- mark rollout step 8 complete

## Root cause

The contest and profile list handlers manually mapped log responses but omitted `duration_seconds`. Several frontend views also assumed every log had amount/unit tracking, leaving duration-only logs blank or showing a misleading amount breakdown.

## Impact

Duration-only logs now render values such as `60 min`, while combined logs render both values, such as `4 pages · 33 min`. Existing amount-only displays remain unchanged.

## Validation

- `bazel test //services/immersion-api/http/rest:rest_test --test_output=errors`
- `pnpm --filter webv2 exec tsc --noEmit`
- `pnpm --filter webv2 lint`
- `pnpm build`
- verified the seeded Listening log exposes `duration_seconds: 3600` through the live dev API

Live browser re-verification was blocked by the dev frontend returning 502 during the final check.


frontend/apps/webv2/app/common/format.ts
```diff
@@ -16,6 +16,38 @@ export const formatScore = (
 export const formatUnit = (amount: number, unit: string) =>
   `${unit.toLowerCase()}${amount !== 1 ? 's' : ''}`
 
+export const formatDuration = (durationSeconds: number) =>
+  `${new Intl.NumberFormat('en-US', {
```
antonve: We shouldn't recreate the formatter every time

PR #709, opened 2026-07-25
Add domain scoring engine

antonve:
## Summary

- document the staged scoring-rules rollout and mark the domain-engine slice complete
- add a storage-independent scoring evaluator with ordered activity, unit, language, tag, and score-source matching
- support amount precedence, duration-minute conversion, zero-score misses, and applied-rule provenance
- support contest `override` fallback and `replace` behavior
- validate scoring inputs and rule-set configuration
- add comprehensive domain tests for evaluation semantics and pinned-version stability

## Why

Scoring currently depends on unit modifiers and hard-coded duration rates. This introduces the pure domain primitive needed for versioned platform and contest rule sets while keeping existing log writes unchanged until later rollout slices add storage and shadow-mode integration.

## Impact

This is an independently deployable domain-only slice. It does not change authoritative scoring, historical scores, API behavior, or the frontend.

## Validation

- `bazel test //services/immersion-api/domain:domain_test`
- `bazel test //services/...` (13 test targets passed)
- `git diff --check`


services/immersion-api/domain/scoring.go
```diff
@@ -0,0 +1,197 @@
...
+	DurationSeconds *int32
+}
+
+type ScoringResult struct {
+	Score            float32
+	ScoreSource      ScoreSource
+	Matched          bool
+	AppliedRuleSetID *uuid.UUID
+	AppliedRuleID    *uuid.UUID
+	AppliedRate      *float32
```
antonve: There could be matches for multiple rules that all should be multiplied together

docs/wip/scoring-rules/README.md
```diff
@@ -34,6 +39,7 @@ For example:
 ```text
 amount: 200 pages * 1.0 = 200
 duration_minutes: 30 minutes * 0.4 = 12
+dense listening: 30 minutes * 0.5 base * 1.4 modifier = 21
```
antonve: listening base should be 0.4, not 0.5. that means the dense modifier should actually be 1.5

services/immersion-api/domain/scoring.go
```diff
@@ -87,26 +88,42 @@ func EvaluateScoringRuleSet(input ScoringInput, ruleSet ScoringRuleSet) (Scoring
 		tags[tag] = struct{}{}
 	}
 
-	for _, rule := range rules {
-		if !scoringRuleMatches(rule, input, source, tags) {
+	var baseRule *ScoringRule
+	modifierRules := make([]ScoringRule, 0)
+	for i := range rules {
+		rule := &rules[i]
```
antonve: question: is this still required in modern golang?

PR #718, opened 2026-07-26
Add guarded database migration recovery workflow

antonve:
## What changed

- Add a separate `/migrate-recovery` binary while keeping automated `/migrate ... up` unchanged.
- Support read-only `inspect` and guarded metadata-only `force`; do not expose `down`, `drop`, or arbitrary steps.
- Require the observed dirty version, a target version no newer than it, and a duplicate target confirmation before forcing.
- Refuse force against clean or unversioned databases, and re-read metadata after force to verify the target is clean.
- Embed the recovery binary in the content, profile, authz, and immersion API images without changing their normal entrypoints.
- Add an operational runbook covering containment, evidence, backup/PITR, physical schema classification, reviewed repair SQL, guarded metadata repair, forward fixes, resumption, and transactional migration requirements.

## Why

The automated migration wrapper intentionally accepts only `up`, which is safe for Argo CD PreSync execution but left dirty or partially applied migration recovery dependent on an improvised external tool. This adds a constrained, image-pinned recovery path without turning the automated runner into a general destructive migration CLI.

`force` still repairs metadata only. The runbook requires operators to verify or repair the physical schema first, and to restore from backup/PITR whenever data integrity is uncertain.

## Validation

- `bazel build //services/...` passed.
- `bazel test //services/...` passed: 15/15 tests.
- `bazel run //:gazelle -- -mode=diff` passed.
- `pnpm --dir docs typecheck` passed.
- OCI audit confirmed `/migrate-recovery` is executable in the layer and all four API entrypoints remain unchanged.
- Disposable PostgreSQL 17.5 integration test passed:
  - produced a real partial migration and `version=2 dirty=true`;
  - verified the partially committed table and row;
  - rejected a force with an incorrect expected dirty version;
  - manually completed the verified physical schema;
  - forced version 2 clean with all guards and post-force verification;
  - applied a forward migration to version 3 clean;
  - repeated `up` as `no change`;
  - rejected force against the clean database.

## Known documentation build issue

The Docusaurus production build fails before document compilation because the existing dependency graph resolves incompatible ProgressPlugin/Webpack options. Retrying from the committed package lock produced the same tooling error. The sidebar TypeScript check passes; this PR does not modify those dependencies.

### DIFF
--- a/docs/docs/database-migration-recovery.md
+++ b/docs/docs/database-migration-recovery.md
@@ -1,0 +1,220 @@
+---
+title: Database migration recovery
+---
+
+# Database migration recovery
+
+This runbook applies when an API image's Argo CD `PreSync` migration Job fails
+and `schema_migrations.dirty` is `true`.
+
+The normal `/migrate` command deliberately supports only forward `up`
+migrations. The separate `/migrate-recovery` command supports:
+
+- `inspect`, which reads the current migration version and dirty state;
+- `force`, which changes migration metadata without running SQL.
+
+It does not support `down`, `drop`, or arbitrary migration steps.
+
+## Safety rules
+
+1. Never use `force` to make an error disappear. It changes only
+   `schema_migrations`; it does not repair tables or data.
+2. Stop automated retries before inspecting or repairing the database.
+3. Ensure exactly one migration or recovery process can access the affected
+   database.
+4. Record the failed image digest, migration version, Job logs, Git revision,
+   database state, and recovery commands in the incident.
+5. Take and verify a database backup or PITR recovery point before manual SQL
+   or metadata repair.
+6. Have a second engineer review the physical schema assessment, repair SQL,
+   and selected target version.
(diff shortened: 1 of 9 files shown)


docs/docs/database-migration-recovery.md
```diff
@@ -0,0 +1,220 @@
...
+1. Never use `force` to make an error disappear. It changes only
+   `schema_migrations`; it does not repair tables or data.
+2. Stop automated retries before inspecting or repairing the database.
+3. Ensure exactly one migration or recovery process can access the affected
+   database.
+4. Record the failed image digest, migration version, Job logs, Git revision,
+   database state, and recovery commands in the incident.
+5. Take and verify a database backup or PITR recovery point before manual SQL
+   or metadata repair.
+6. Have a second engineer review the physical schema assessment, repair SQL,
```
antonve, line 29: Just say human, I'm the only dev on this project. The executor is most likely an AI agent.

PR #744, opened 2026-08-02
Switch platform logs to the scoring engine

antonve:
Stacked on #743.

## What changed

Make the active platform rule set authoritative for log create and update when API_SCORING_ENGINE_ENABLED is true. The flag defaults to false and the interim scorer remains the fallback while disabled.

## Why

This introduces the production cutover as an explicit, reversible deployment decision after shadow parity is verified.

## Validation

- full Bazel service build and all 16 backend test targets
- full pnpm frontend build, WebV2 typecheck, and lint

### DIFF
--- a/services/immersion-api/deployments/api.yaml
+++ b/services/immersion-api/deployments/api.yaml
@@ -54,6 +54,8 @@ spec:
             value: "redis://valkey-immersion.default:6379"
           - name: SERVICE_NAME
             value: "immersion-api"
+          - name: API_SCORING_ENGINE_ENABLED
+            value: "false"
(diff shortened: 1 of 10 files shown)


services/immersion-api/deployments/api.yaml
```diff
@@ -54,6 +54,8 @@ spec:
             value: "redis://valkey-immersion.default:6379"
           - name: SERVICE_NAME
             value: "immersion-api"
+          - name: API_SCORING_ENGINE_ENABLED
```
antonve, line 57: Does this flag default to false in the code?

PR #748, opened 2026-08-02
Add rule set management API

antonve:
Stacked on #747.

## What changed

Add platform and contest endpoints to list rule sets, create drafts, publish immutable versions, and activate a published set. Platform mutations require admin access; contest mutations require ownership and are locked after the contest starts.

## Why

Rule changes need a validated, versioned management workflow rather than direct database edits.

## Validation

- domain, repository, and REST handler coverage
- full Bazel service build and all 16 backend test targets
- full pnpm frontend build, WebV2 typecheck, and lint


services/immersion-api/http/rest/openapi/api.yaml
```diff
@@ -936,6 +936,124 @@ paths:
...
+            schema:
+              $ref: "#/components/schemas/ScoringRuleSetDraft"
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ScoringRuleSet"
+  /contests/{id}/scoring-rule-sets:
```
antonve: Should be `/contests/{id}/scoring/rule-sets`

PR #753, opened 2026-08-02
Add scoring rule schema migration

antonve:
## Summary

Adds only migration `0021_scoring_rule_storage`, which creates and seeds the scoring rule schema.

This is intentionally isolated from all generated storage code and runtime behavior so it can be reviewed, merged, and deployed independently. The schema uses declarative constraints only; application-defined database functions and triggers are not used. Seed rows receive generated UUIDs rather than hardcoded identifiers.

## Deployment order

1. Merge this PR to `main`.
2. Deploy and verify the migration.
3. Continue with the next code slice in the stack.

## Validation

- `git diff --check`
- applied migrations `0001` through `0021` to a clean PostgreSQL 16 database
- rolled back `0021` and verified its tables were removed
- `bazel test //services/...` (16 test targets)

### DIFF
--- a/services/immersion-api/storage/postgres/migrations/0021_scoring_rule_storage.up.sql
+++ b/services/immersion-api/storage/postgres/migrations/0021_scoring_rule_storage.up.sql
@@ -1,0 +1,213 @@
+begin;
+
+create table scoring_rule_sets (
+  id uuid primary key default uuid_generate_v4(),
+  scope text not null,
+  contest_id uuid references contests(id),
+  version integer not null,
+  status text not null,
+  mode text,
+  fallback_rule_set_id uuid references scoring_rule_sets(id),
+  created_at timestamp not null default now(),
+  published_at timestamp,
+
+  constraint scoring_rule_sets_scope_valid
+    check (scope in ('platform', 'contest')),
+  constraint scoring_rule_sets_version_positive
+    check (version > 0),
+  constraint scoring_rule_sets_status_valid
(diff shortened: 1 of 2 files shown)


services/immersion-api/storage/postgres/migrations/0021_scoring_rule_storage.up.sql
```diff
@@ -0,0 +1,328 @@
...
+  constraint scoring_rules_score_source_valid
+    check (score_source in ('amount', 'duration_minutes')),
+  constraint scoring_rules_rate_valid
+    check (rate >= 0 and rate < 'Infinity'::real)
+);
+
+create unique index scoring_rules_rule_set_priority
+  on scoring_rules(rule_set_id, priority);
+
+create function validate_scoring_rule_set_fallback()
```
antonve: This should not exist, this should be application business logic. This applies to all functions and triggers

services/immersion-api/storage/postgres/migrations/0021_scoring_rule_storage.up.sql
```diff
@@ -0,0 +1,328 @@
...
+before insert or update of scoring_rule_set_id on contests
+for each row execute function validate_contest_scoring_rule_set();
+
+insert into scoring_rule_sets (
+  id,
+  scope,
+  version,
+  status
+) values (
+  '10000000-0000-0000-0000-000000000001',
```
antonve: We should not be hardcoding ids in the migrations

PR #812, opened 2026-08-21
Make the Tilt dev stack deterministic

antonve:
## Summary

- require explicit durable Tilt configuration and allow the `homelab-dev` context
- make teardown work with operator-managed secrets and use anonymous registry auth for frontend builds
- gate every database consumer on Zalando Postgres readiness and propagate operator credentials into API namespaces
- run Kratos/Keto migrations as init containers, start pgweb only after Postgres, and remove hardcoded database passwords
- restart the actual database consumers during `dev-reset`
- keep immersion seed rows compatible with stable unit-key constraints

## Validation

- evaluated the root Tiltfile for both `homelab-dev` and `orbstack`
- rendered and client-validated the Kratos/Keto charts, pgweb, and all four API manifests
- ran a clean live `homelab-dev` startup: Postgres, Kratos, Keto, pgweb, all frontends, and all four APIs are healthy
- verified Kratos/Keto migrations completed as init containers and no migration Jobs remain
- verified pgweb connected using a passfile and does not retain `PGPASSWORD` in its process environment
- verified copied API credentials match the Zalando operator sources and are owned by their target namespaces
- ran `dev-seed` twice successfully and verified seeded base/contest unit keys
- verified `https://tadoku.dev.lab`, account, admin, and pgweb endpoints
- ran shell syntax checks and `git diff --check`

### DIFF
--- a/frontend/Tiltfile
+++ b/frontend/Tiltfile
@@ -6,6 +6,7 @@ load(
   'is_shared_cluster',
   'render_template',
 )
+load('./../infra/dev/build/Tiltfile', 'anonymous_docker_config_prefix')
 
 def _sh_quote(value):
   return "'" + value.replace("'", "'\"'\"'") + "'"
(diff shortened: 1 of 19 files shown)


frontend/Tiltfile
```diff
@@ -19,15 +19,17 @@ def _buildx_output():
 def _buildx_setup_command():
   builder = _sh_quote(dev_buildx_builder())
   return (
-    'docker buildx inspect {builder} >/dev/null 2>&1 || ' +
-    'docker buildx create --name {builder} --driver docker-container >/dev/null; ' +
-    'docker buildx inspect --bootstrap {builder} >/dev/null; '
-  ).format(builder=builder)
+    'TADOKU_FRONTEND_DOCKER_CONFIG="${XDG_CACHE_HOME:-/tmp}/tadoku/frontend-buildx-anonymous"; ' +
+    'mkdir -p "$TADOKU_FRONTEND_DOCKER_CONFIG"; ' +
```
antonve: why is this frontend specific? doesn't bazel go builds also use this?

scripts/dev/sync-db-secrets.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
antonve, line 1: what's the reason for requiring this script?

PR #823, opened 2026-08-23
Limit display names to 32 characters

### LINKED ISSUE #94: Limit display name length

antonve:
## Summary

- enforce a 32-character display-name maximum in the Kratos identity schema
- hardcode matching 2/32-character bounds on the shared registration and settings input

## Verification

- pnpm --filter auth lint
- pnpm --filter auth exec tsc --noEmit
- pnpm --filter auth build

Fixes #94


frontend/apps/auth/src/ui/displayName.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import Ajv from 'ajv'
+import { parse } from 'yaml'
+import { describe, expect, it } from 'vitest'
+import {
+  DISPLAY_NAME_MAX_LENGTH,
+  inputMaxLength,
+} from './displayName'
+
+const kratosValuesPath = new URL(
+  '../../../../../infra/dev/ory/kratos_values.yaml',
```
antonve: Wtf? This file is not used in production, we shouldn't include these files. In the real image we cant even access this path

antonve (commented):
This solution is way too over engineered, just put a hardcoded username min max. No need for tests

frontend/apps/auth/src/ui/displayName.test.ts
antonve: Simplified as requested: removed the test harness, helper, and test-only dependencies. The production diff is now limited to the Kratos maxLength and direct 2/32 bounds on the shared auth input.

**Posted by gpt-5.6-sol**

PR #829, opened 2026-08-30
Test log submission against the migrated schema

antonve:
## Summary

- add a reusable PostgreSQL test helper that creates an isolated database and applies every checked-in immersion migration through the real golang-migrate runner
- execute the log-submission domain flow with the real repository, covering the database commit followed by the response lookup
- verify the returned log and independently confirm exactly one row was persisted
- provide PostgreSQL 17 to the Bazel CI job so this regression coverage cannot silently skip there

## Why

Migration 0025 added users.deleted_at, which made the post-commit FindLogByID response query ambiguous. A submission persisted successfully but returned HTTP 500, inviting retries and duplicate logs. sqlc generation did not detect that runtime failure.

## Validation

- bazel run //:gazelle -- -mode=diff
- bazel build //services/immersion-api/...
- IMMERSION_TEST_POSTGRES_URL=postgres://... bazel test //services/immersion-api/... --test_output=errors --nocache_test_results
- negative proof: restoring the pre-#828 unqualified FindLogByID predicate makes the migrated-schema submission test fail with pq: column reference "deleted_at" is ambiguous

## Follow-up

The test includes a TODO to cover this through the public POST /logs endpoint so authentication, request mapping, commit, and response rendering are eventually exercised together.

After this lands, #827 can replace its minimal hand-created users schema with the shared postgrestest.OpenMigratedDatabase helper and use the common IMMERSION_TEST_POSTGRES_URL CI environment.

**Posted by gpt-5.6-sol**


services/immersion-api/storage/postgres/postgrestest/database.go
```diff
@@ -0,0 +1,100 @@
...
+			require.NoError(t, testDB.Close())
+		}
+		_, dropErr := adminDB.Exec(`drop database "` + databaseName + `" with (force)`)
+		require.NoError(t, dropErr)
+		require.NoError(t, adminDB.Close())
+	})
+	require.NoError(t, err)
+	require.NoError(t, testDB.Ping())
+
+	migrationDirectory := filepath.Dir(migrationPaths(t)[0])
```
antonve: seems weird to only grab the first path?

services/immersion-api/storage/postgres/migrated_queries_test.go
```diff
@@ -0,0 +1,151 @@
+package postgres_test
```
antonve: Leave todo that we should be using e2e tests in the future

antonve: and actually, the biggest failure was in submitting logs, we might wanna cover that?

services/immersion-api/storage/postgres/postgrestest/database.go
antonve: Agreed. I refactored the helper to resolve and return the migration directory directly. It still enumerates every `.up.sql` file to validate that the migration sequence is complete and gap-free. Addressed in 4d81f525.

**Posted by gpt-5.6-sol**

services/immersion-api/storage/postgres/migrated_queries_test.go
antonve: Added a TODO calling for coverage through the public `POST /logs` endpoint so authentication, request mapping, commit, and response rendering are eventually exercised together. Addressed in 4d81f525.

**Posted by gpt-5.6-sol**

antonve: Covered the dangerous submission path in 4d81f525. The new migrated-schema integration test runs `LogCreate.Execute` with the real PostgreSQL repository, which commits the write and then performs the response lookup; it verifies the returned log and independently confirms exactly one persisted row. Full HTTP coverage remains captured by the TODO above.

**Posted by gpt-5.6-sol**

services/immersion-api/storage/postgres/postgrestest/database.go
```diff
@@ -0,0 +1,101 @@
...
+	migrationErr := migrator.Up()
+	sourceCloseErr, databaseCloseErr := migrator.Close()
+	require.NoError(t, migrationErr)
+	require.NoError(t, sourceCloseErr)
+	require.NoError(t, databaseCloseErr)
+
+	return testDB
+}
+
+func migrationDirectory(t testing.TB) string {
```
antonve: this is a bad name if it does more than just returning the migration dir

services/immersion-api/storage/postgres/migrated_queries_test.go
antonve: let's delete this file and just keep the logcreate test

services/immersion-api/storage/postgres/postgrestest/database.go
antonve: Agreed. Renamed it to `validatedMigrationDirectory` so the sequence-validation side effect is explicit. Addressed in fad090a5.

**Posted by gpt-5.6-sol**

services/immersion-api/storage/postgres/migrated_queries_test.go
antonve: Done in fad090a5. The four-query test and fixture are removed; the PR now focuses on the migrated-schema log-submission test. I also updated the PR title and description to match the narrower scope.

**Posted by gpt-5.6-sol**

PR #835, opened 2026-09-01
Gate the V2 log journey with Flipt

antonve:
## Summary

- use the Flipt decision as the sole source of access to the existing V2 create, details, and edit log journey, including for admins
- keep the legacy journey as the safe default and latch the selected journey for each mounted page so refreshes cannot discard in-progress form input
- bootstrap SSR decisions through a server-only internal Oathkeeper endpoint, while browser refreshes continue using the public endpoint
- keep backend behavior unchanged because V1 and V2 already use compatible authenticated endpoints and the scoring-engine switch is an independent concern

## Safety

- missing, failed, or disabled decisions resolve to the legacy experience
- no frontend role or admin bypass can diverge from Flipt targeting
- no Flipt SDK, client token, or sensitive configuration is added to the browser
- the dev Flipt seed remains disabled by default; local live-test targeting was restored to admin-only

## Verification

- `pnpm --filter webv2 test` — 31 tests passed
- `pnpm --filter webv2 exec tsc --noEmit`
- `pnpm --filter webv2 lint`
- `pnpm build`
- standard root `tilt up` healthy for webv2, Oathkeeper, Flipt, and immersion-api
- live reader decision verified through both API and SSR with targeting enabled, then verified legacy SSR after targeting was removed

### DIFF
--- a/frontend/apps/webv2/pages/logs/[id]/edit.tsx
+++ b/frontend/apps/webv2/pages/logs/[id]/edit.tsx
@@ -6,21 +6,33 @@ import { EditLogForm } from '@app/immersion/EditLogForm/Form'
 import { useRouter } from 'next/router'
 import Head from 'next/head'
 import { useSessionOrRedirect, useUserRole } from '@app/common/session'
+import { useLatchedFeatureFlag } from '@app/feature-flags/client'
 
 const Page = () => {
   const router = useRouter()
   const id = router.query['id']?.toString() ?? ''
   const role = useUserRole()
+  const useV2 = useLatchedFeatureFlag(
+    'release-log-entry-v2',
+    role !== undefined,
+    role === 'admin',
+  )
   const log = useLog(id, { enabled: !!id })
   const options = useLogConfigurationOptions()
 
   useSessionOrRedirect()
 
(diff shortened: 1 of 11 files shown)


frontend/apps/webv2/pages/logs/[id]/edit.tsx
```diff
@@ -6,21 +6,33 @@ import { EditLogForm } from '@app/immersion/EditLogForm/Form'
...
+import { useLatchedFeatureFlag } from '@app/feature-flags/client'
 
 const Page = () => {
   const router = useRouter()
   const id = router.query['id']?.toString() ?? ''
   const role = useUserRole()
+  const useV2 = useLatchedFeatureFlag(
+    'release-log-entry-v2',
+    role !== undefined,
+    role === 'admin',
```
Comment (antonve) [inline frontend/apps/webv2/pages/logs/[id]/edit.tsx]: remove admin check, we should define who has access in flipt

Comment (antonve) [reply frontend/apps/webv2/pages/logs/[id]/edit.tsx]: Removed the admin bypass from the shared latch and all three log routes. Flipt is now the sole V2 access decision for every user. I also removed the override API and its tests so the role-based path cannot be reintroduced accidentally here.

Verified with the full webv2 test suite (31 passing), typecheck, lint, and production build.

**Posted by gpt-5.6-sol**

PR #856, opened 2026-09-05
Use Tadoku API as the default development gateway upstream

antonve:
The existing Tadoku API proxy is built but is not included in Tilt. This adds the service and makes it the default gateway upstream for the four browser API prefixes and the internal feature-flag request in both local and shared development. The dev ingress waits for the facade to be ready.

Oathkeeper owns the external `/api/internal` prefix and strips it before forwarding. For example, the browser requests `/api/internal/content/pages/blog`, Tadoku API receives `/content/pages/blog`, and its legacy Content proxy forwards `/pages/blog`. Native handlers can gradually replace the domain proxy routes without depending on the external prefix or changing browser URLs.

Routing uses fixed upstreams in the existing access rules. Authentication, service-token exchanges, and the Flipt authorization callback retain their existing configuration.

Go HTTP tests compare direct and proxied requests using the real legacy JWT/identity/audience middleware (24 scenarios), preserve bodies and headers across 12 response statuses, and verify streaming. Tests use the domain paths and confirm browser-prefixed paths are not registered directly in Tadoku API. A loopback benchmark supplies a repeatable direct/proxy baseline.

Validation: `bazel build //services/...`, `bazel test //services/...`, and `git diff --check` passed. Gazelle output was verified when the test file was added; no imports or source files changed in the prefix revision. Database-dependent tests use their existing environment-gated behavior locally; CI provides PostgreSQL. The initial loopback sample was 21,727 ns/op direct and 57,105 ns/op proxy; it predates the prefix revision and does not represent deployed latency.

Full-stack Oathkeeper/Kratos/application contract rehearsal and deployed latency measurements remain pending before production activation. Upstash setup is independent of this development work.

[Round scope and remaining rehearsal work](https://draft.apps.lab/documents/tadoku-api-facade-dev-round-2026-09.html?version=3).

**Posted by gpt-6-astra**


infra/dev/ory/access_rules.yaml
```diff
@@ -1,8 +1,7 @@
 - id: "tadoku:content:api"
   upstream:
     preserve_host: true
-    url: "http://content-api.tdk-content-api:80"
-    strip_path: /api/internal/content/
```
antonve, line 5: So one thing that I want to take into account is that I don't want the API slash internal thing in the real API implementation inside the Tadoq API. So I do wonder if this is the right configuration here, keeping in mind that we need to gradually translate things over.

antonve: Agreed. Oathkeeper should own `/api/internal`. Updated the five gateway rules to strip only that prefix, and Tadoku API now registers `/content/`, `/authz/`, `/immersion/`, and `/profile/`.

The path flow is now `/api/internal/content/pages/blog` → `/content/pages/blog` in Tadoku API → `/pages/blog` in the legacy Content service. As endpoints move natively, their handlers can take over the domain paths while the remaining paths continue through the proxy. They never need the external `/api/internal` prefix.

Updated the Go HTTP tests to use those domain paths and added a check that browser-prefixed paths are not registered directly in Tadoku API. Backend build and tests pass.

**Posted by gpt-6-astra**

PR #868, opened 2026-09-13
test(tadoku-api): verify announcements parity against legacy Content API

antonve:
## Summary

- Run the six existing announcements-list SQL/request/golden fixtures against both native Tadoku API and legacy Content API: 12 full-response comparisons, with no golden changes.
- Reuse the real production route registration, handlers, domain operations, repositories and generated queries in process. Construct both handlers once in `TestMain`; reset/reseed before each implementation and close both pools at teardown.
- Keep legacy dependencies test-only. Update the integration-test import policy and testing guidance; leave runtime code, generated code, schema and the deferred DB helpers untouched.

## Clock control and scope

Legacy SQL reads PostgreSQL `now()`; the native operation uses `timex.Now()`. A small test-only connection wrapper binds the legacy query's two clock calls to the same frozen scenario time. It rejects unexpected query names, clock-call counts and argument counts. Filtering, ordering, limits and mapping still run against real PostgreSQL; HTTP responses are compared exactly, with only the existing line-ending normalization.

Tests also verify that changing the scenario clock affects the reused legacy connection and that unsupported query shapes fail. This is announcements-list endpoint parity, not parity for all API operations, authentication/gateway middleware, or the legacy choice of database clock. Other operations remain proxied and can gain shared golden cases as they migrate.

## Verification

Using disposable loopback PostgreSQL 17 and local Bazel caching:

- `bazel test //services/... --test_output=errors` — all 207 backend targets build; all 36 test targets pass.
- `bazel test //services/tadoku-api/... --@rules_go//go/config:race --test_output=errors` — all eight test targets pass with the race detector.
- `bazel test //services/tadoku-api/e2e:e2e_test --test_arg=-test.count=10 --test_arg=-test.shuffle=on --test_output=errors` — ten shuffled suite runs pass.
- `bazel run //tools/ci/depolicy` and `bazel run //:gazelle -- -mode=diff` — pass.
- Bazel dependency query — no legacy service or Echo dependency in the native binary.
- `gofmt` and `git diff --check` — clean.

The existing CI E2E/race targets include the new comparisons without workflow changes. No deployment or migration is included.


services/tadoku-api/README.md
```diff
@@ -54,8 +54,9 @@ the contract's `x-tadoku-operational-surfaces` metadata, not rendered as product
 The native read returns the existing announcement JSON shape, including nullable
 hrefs and an empty array when nothing is active. Publication windows, soft deletion,
 namespace filtering, descending order and the ten-item limit remain unchanged.
-Database failures return 500 without falling back to the proxy. There is no legacy
-server comparison or duplicated authentication matrix in its tests.
+Database failures return 500 without falling back to the proxy. The native and
+legacy announcement handlers share the same request/response golden tests;
```
antonve: Generalize this

services/tadoku-api/README.md
```diff
@@ -145,6 +147,35 @@ There is no automatic recording mode: edit and review the expected files for an
 intentional contract change. Lifecycle tests stay focused on startup/shutdown,
 not a growing list of endpoint assertions.
 
+### Legacy parity
+
+The six announcements cases each run as `tadoku-api` and `content-api` subtests.
```
antonve: Generalize these statements, we are working on a full application port. Stop saying "this specific portion does x" when it is a statement that should apply generally

antonve: Encode this in agents.md

services/tadoku-api/README.md
antonve: Updated in a7acc0a1. The compatibility section now requires every migrated operation to preserve its API inputs, outputs and business behavior, with shared native/legacy golden cases and no proxy fallback on native failures. It no longer describes that contract as announcement-specific.

**Posted by gpt-6-astra**

services/tadoku-api/README.md
antonve: Updated in a7acc0a1, including the follow-up about AGENTS.md. The README now describes parity, fixture reuse, test lifecycle and test-only clock handling as application-wide conventions. Actual coverage and the existing clock-binding limit are labeled separately rather than presented as the scope of those rules.

AGENTS.md now explicitly requires architecture, compatibility and testing guidance to apply to the full application port, with examples/current coverage/limitations kept separate. Its parity rule also explicitly applies to every migrated operation.

Documentation-only revision; git diff --check passed. No test or production code changed.

**Posted by gpt-6-astra**

antonve: Updated in b565a2de: removed the coverage sections and endpoint-specific walkthrough/examples from the README, rather than retaining them under separate labels. AGENTS.md now prohibits endpoint implementation/coverage sections in general documentation.

Added services/tadoku-api/MIGRATION_LOG.md, linked from the README, with an unchecked follow-up to remove all references to the “native” API and use “Tadoku API” instead. Documentation-only; diff checks passed.

**Posted by gpt-6-astra**

PR #870, opened 2026-09-14
feat(tadoku-api): add Keto ban enforcement and golden parity tests

antonve:
## Summary

- Add read-only Keto integration and a single shared timeout → authentication → ban-check chain for all application routes. Normal `Handle` / `HandleFunc` registration inherits this chain automatically. No roles management or business-logic changes; probes and proxied routes keep their existing boundaries.
- Add a reusable, test-only Keto fixture: Bazel downloads the official SHA256-pinned v25.4.0 SQLite-enabled executable; the suite owns an in-memory process on random loopback ports using the checked-in OPL model. No separately running Keto or Docker-based Keto is needed.
- Keep optional `relationships.json` beside each scenario's SQL and HTTP fixtures. Shared setup resets both stores before each implementation; Keto cleanup removes all relationships, including writes made during a scenario.
- Compare six shared HTTP goldens against the real old and new authentication/ban middleware stacks. Middleware test endpoints are registered normally on the production router and end at test-only success handlers, without business logic. Keep provider failure/cancellation checks separate.

## Compatibility and rollout

- Banned users receive an empty 403; being an administrator does not override a ban. Guests and missing/empty identities skip the lookup. Ban-check errors are logged and fail open, preserving the existing policy.
- The new gate makes one ban lookup, not the legacy admin-plus-ban lookup. An unrelated admin-check failure therefore no longer discards a successful ban result. All six relationship golden cases match legacy; none are skipped.
- Runtime requires `API_KETO_READ_URL` only. The dev manifest is updated; production configuration must provide it before deployment. This PR does not deploy anything or configure a write credential.
- The disposable Keto executable currently supports Linux x86-64. PostgreSQL test prerequisites are unchanged.
- No database migrations, deferred DB/repository helper changes, or production permission mutations are included.

## Verification

Implementation was delegated to GPT-5.6 Sol with high reasoning and independently reviewed in the main thread. Main-thread verification passed:

- `bazel build --lockfile_mode=error //services/...` — 210 targets.
- `bazel test --lockfile_mode=error //services/...` — all 37 test targets.
- Latest review revision: race detector with shuffled order and ten repetitions — HTTP, E2E, startup, and testketo all passed. The initial integration also passed all nine relevant race targets.
- `bazel run --lockfile_mode=error //:gazelle -- -mode=diff` — no differences.
- `bazel run --lockfile_mode=error //tools/ci/depolicy` — 45 Go files, including 20 test files; zero violations.
- Runtime dependency query — no legacy API/authz/middleware, Echo, or Keto test executable in the Tadoku API binary graph.
- Go formatting and non-HTTP whitespace checks pass. Golden files intentionally preserve empty header values and HTTP message terminators.

The initial commits separate the reusable Keto test infrastructure from middleware integration and parity coverage. Review revisions separately rename the router fields, extract startup's Keto middleware constructor, and simplify the middleware fixtures and suite lifecycle.

## Review correction: protected route registration

The follow-up commit composes middleware once around a private application router. Every normal route registration points through that shared handler; the private outer router keeps probe and temporary proxy exceptions explicit, including legacy HEAD ownership. This uses the standard library with no new routing dependency.

The new regression failed before the fix (both added routes bypassed authentication and bans) and passes afterward. It verifies shared composition, timeout propagation, authentication-before-ban ordering, short-circuiting, and escaped path values. Main-thread revalidation: all 210 build targets and 37 test targets passed with test caching disabled for the tests; HTTP, E2E, and startup suites also passed ten shuffled repetitions under the race detector. Gazelle and Depolicy passed. Existing golden responses are unchanged.

The latest review also moves middleware goldens onto normal test-only router registrations, removes the redundant business-endpoint wiring tests and duplicated ban-policy unit cases, makes skipped test gates and hostile parent environment settings explicit, and documents offline synthetic JWT authoring. Resource-local defers preserve reverse-order cleanup and test exit status. A deliberately invalid PostgreSQL setup failed as expected after Keto startup and left no Keto process behind; the normal E2E suite passed afterward. The method/path routing regression failed against detached handlers before the fix and passes now. Existing golden requests/responses and JWKS remain unchanged.


services/tadoku-api/cmd/tadoku-api/main.go
```diff
@@ -123,8 +129,12 @@ func start(cfg config, logger *slog.Logger) (*application, error) {
 	contentRepository := content.NewAnnouncementsRepository(pool)
 	contentService := content.NewService(contentRepository)
 	api := app.New(contentService)
+	keto := ketoclient.NewReadClient(cfg.KetoReadURL)
+	rejectBanned := transporthttp.RejectBannedUsers(func(ctx context.Context, subjectID string) (bool, error) {
```
antonve: extract into function

services/tadoku-api/e2e/testdata/BannedUsers/200_admin/request.http
```diff
@@ -0,0 +1,4 @@
+GET /test/banned HTTP/1.1
+Host: api.test
+Authorization: Bearer eyJhbGciOiJSUzI1NiIsImtpZCI6ImF1dGhlbnRpY2F0aW9uLWZpeHR1cmUifQ.eyJpc3MiOiJodHRwOi8vb2F0aGtlZXBlci1hcGkvIiwic3ViIjoiMTExMTExMTEtMTExMS00MTExLTgxMTEtMTExMTExMTExMTExIiwiaWF0IjoxNzg5MjE0MzQwLCJuYmYiOjE3ODkyMTQzNDAsImV4cCI6MTc4OTIxNDQ2MCwidHlwZSI6InVzZXIiLCJzZXNzaW9uIjp7ImlkZW50aXR5Ijp7InRyYWl0cyI6eyJkaXNwbGF5X25hbWUiOiJSZWFkZXIgT25lIiwiZW1haWwiOiJyZWFkZXJAZXhhbXBsZS50ZXN0In19fX0.G5cQUa7a3x4k1Am6xLOwwJlgXjtIMPD4tsY6MkLFwPyPciUosa5etO0R2T7Rp6YOzLxUupkG3jnfTLWv2KdGpxGiO8gjBAD7Nrx87mFECQ29DHOd6waGYcqM9fCqjVgjFAxo_NwbP7EfdWD9_dt_ycpJfEv-FaEMwBT7R49jWWIDPZAgYRtXmr0Tlc5nclSVd6b5SsizbxZFb0WZGlhzkKNUwetiiXAYlZPjuU4p90wqe1QnC1LZXJv6izfDCeUU75aibQUPkz16RLjQL4E0FroSaFw97AzBvFYx4lWqa4wjHN0_Z0UiEIrkjGpEf9MrZ1dUotBBsb2iw3nyq_e0XQ
```
antonve, line 3: we probably need some documentation on how these tokens for the test cases are generated (so we can do it without LLMs as well)

services/tadoku-api/e2e/api_test.go
```diff
@@ -41,18 +46,23 @@ func runTests(m *testing.M) (code int) {
...
+	keto, err = testketo.New(context.Background())
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		return 1
+	}
 
 	api, err = newTestAPI(context.Background())
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		fmt.Fprintln(os.Stderr, errors.Join(err, keto.Close()))
```
antonve: I wonder if there is a better way of handling this kind of setup code? doesn't seem to scale well

services/tadoku-api/e2e/banned_users_test.go
```diff
@@ -0,0 +1,134 @@
...
+			{name: "banned", relationships: filepath.Join(path, "relationships.json"), want: http.StatusForbidden},
+			{name: "allowed", want: http.StatusOK},
+		} {
+			t.Run(test.name, func(t *testing.T) {
+				if test.relationships == "" {
+					reset(t)
+				} else {
+					reset(t, test.relationships)
+				}
+				request := httptest.NewRequest(http.MethodGet, "/content/announcements/main/active", nil)
```
antonve: Why are we testing with a route like this? I thought our instructions in the docs and such imply that we should use normal routes for middleware testing?

services/tadoku-api/e2e/routing_test.go
```diff
@@ -24,13 +24,13 @@ func TestRouterWorksWithoutLegacyProxyRoutes(t *testing.T) {
 	application := app.New(service)
 	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
 
-	handler, err := transport.NewHandler(application, api.db.Pool.Ping, time.Second, logger, withoutAuthentication)
+	handler, err := transport.NewHandler(application, api.db.Pool.Ping, time.Second, logger, withoutAuthentication, withoutAuthentication)
```
antonve: this reads really sus, i know it's for banned users testing but the code here is misleading

services/tadoku-api/internal/testketo/keto_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"errors"
+	"net/http"
+	"net/url"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestResetSeedsAndClearsEveryNamespace(t *testing.T) {
+	t.Setenv("DSN", "postgres://127.0.0.1:1/unsafe")
```
antonve, line 17: why is this postgres? keto is supposed to run in memory

services/tadoku-api/transport/http/banned_users_test.go
```diff
@@ -0,0 +1,125 @@
...
+	stdhttp "net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/identity"
+)
+
+func TestRejectBannedUsers(t *testing.T) {
```
antonve: isn't this pretty much a duplicate of the e2e test?

services/tadoku-api/cmd/tadoku-api/main.go
antonve: Extracted the Keto client and ban-check binding into `newBannedUserMiddleware`. Startup now just constructs that middleware and passes it to the router; the check and request-context propagation are unchanged.

Implemented in [17b18a51](https://github.com/tadoku/tadoku/commit/17b18a51).

**Posted by gpt-6-astra**

services/tadoku-api/e2e/testdata/BannedUsers/200_admin/request.http
antonve: Added a copy-paste offline token-authoring recipe to the README using Node's built-in crypto plus jq, with no new project dependency. It creates a synthetic key, prints the public JWK and a signed token, and explains the frozen clock and how to add the public key without replacing existing fixtures. The private key is never printed or saved. When only Keto relationships differ, the existing signed user token can simply be reused. I verified the recipe and kept the existing requests/goldens unchanged.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

services/tadoku-api/e2e/api_test.go
antonve: Simplified this to resource-local defers registered immediately after each successful setup. Each new dependency now needs its own setup and deferred close, without editing a central cleanup list or earlier error branches. Cleanup runs in reverse acquisition order, joins cleanup errors, and preserves a nonzero test exit status. `newTestAPI` likewise has one partial-setup cleanup path. No lifecycle framework or `testing.TB` dependency added. I also checked an intentionally failed PostgreSQL setup: it exited nonzero and the already-started Keto process was cleaned up.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

services/tadoku-api/e2e/banned_users_test.go
antonve: Removed the announcement-based wiring tests. The authentication and ban goldens now dispatch through the real production router: `/test/authentication` and `/test/banned` are registered with its normal `HandleFunc`, so they inherit the shared middleware automatically. Their handlers only expose the test response/context; no business operation is involved, and neither route exists in production. Added a regression for normal method/path dispatch; it failed against the detached handlers and passes with the router registrations. Updated the testing guidance to match.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

services/tadoku-api/e2e/routing_test.go
antonve: Replaced the misleading repeated `withoutAuthentication` argument with explicit `skipAuthentication` and `skipBanCheck` test-only functions. The accompanying comment explains that this router intentionally isolates endpoint contracts from both shared gates. Middleware goldens use the other suite router with the real gates; production has no opt-out.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

services/tadoku-api/internal/testketo/keto_test.go
antonve: Keto does run in memory (`dsn: memory`). This deliberately unusable PostgreSQL DSN is set only in the parent test process to prove the child does not inherit ambient connection settings. If it leaked into Keto, startup would fail against port 1. Added an explicit comment at the assignment so it cannot be mistaken for the fixture's database configuration.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/banned_users_test.go
antonve: Yes—the ordinary guest/allowed/banned cases duplicated the real-Keto goldens, so I removed them from this unit matrix. The focused tests now cover missing/empty identity guards, provider failures (including error precedence over a `true` result), logging, and request-context/deadline propagation. Policy outcomes and roles remain covered by the shared old/new HTTP goldens.

Implemented in [01900c02](https://github.com/tadoku/tadoku/commit/01900c02).

**Posted by gpt-6-astra**

antonve:
for some weird bug in github this pr was already merged into main but the pr is not recognized as merged?

PR #873, opened 2026-09-15
Migrate admin announcements list to Tadoku API

antonve:
Rebased onto main after #874 merged. Includes the admin announcements-list migration and review follow-ups.

## Change

- Serve `GET /content/announcements/{namespace}` through the generated strict HTTP server, preserving the existing layers: HTTP handler → application → content service → AnnouncementsRepository → generated SQLC.
- Generated code owns route registration, parameter binding and response serialization.
- Check `RequireAdmin` before opening the application transaction. Construct the real checker through `permissions.NewKetoChecker(client)`; the content service owns pagination and repository methods only access/map data.
- Preserve the response envelope, ordering, namespace/deleted-row filtering, pagination defaults/caps, and malformed/negative pagination behavior.
- Keep HEAD and write methods on the legacy proxy.
- No schema migrations, announcement writes, private DB-helper changes, admin middleware, or endpoint-coverage documentation.

## Review follow-ups

- One suite-level Tadoku router with real JWT authentication, ban middleware and Keto-backed permissions. The legacy endpoint router uses its production auth stack too.
- Signed requests and explicit per-case `relationships.json` for admin access; public cases use signed guest JWTs. No passthrough middleware, injected admin identity, or always-allow checker.
- Shared admin/non-admin/missing-credential response goldens run through both APIs. Dependency-failure tests remain isolated, including a real Keto shutdown after first proving the seeded admin can access both implementations.
- Remove the SQL-rewriting connector and its three helper-specific tests. The legacy active-list service now takes its existing `commondomain.Clock` and passes one time value into SQLC. Production therefore uses its UTC application clock rather than database `now()`; comparison boundaries are unchanged.
- Update durable testing conventions to require real auth and normal clock dependencies.

## Verification

- Full backend build: 212 targets passed.
- Full backend tests: all 38 test targets passed.
- Seven focused targets, including HTTP E2Es and the touched legacy domain, passed ten shuffled race-enabled repetitions.
- Existing response goldens remain unchanged; a missing-credential golden was added.
- SQLC regenerated using all pinned package versions; complete generated diff included.
- Gazelle clean; Depolicy: 50 Go files, zero violations.
- Docs/OpenAPI checks passed during the preceding rebase verification; canonical specs and generated HTTP bindings are unchanged by these follow-ups.

Original feature implementation by gpt-5.6-sol (high). Rebase and review follow-ups by gpt-6-astra. No deployment performed.

**Posted by gpt-6-astra**


services/tadoku-api/app/announcements.go
```diff
@@ -4,8 +4,23 @@ import (
...
+func (a *Application) ListAnnouncements(ctx context.Context, namespace string, pageSize, page int) (*content.AnnouncementList, error) {
+	if err := a.permissions.RequireAdmin(ctx); err != nil {
+		return nil, err
+	}
+
+	var result *content.AnnouncementList
+	err := postgres.RunInTransaction(ctx, a.db, func(ctx context.Context) error {
+		var err error
+		result, err = a.content.ListAnnouncements(ctx, namespace, pageSize, page)
+		return err
```
antonve: do we actually need to define err and return err? can't we directly do (err error) in the return type to avoid this?

services/tadoku-api/transport/http/announcements.go
```diff
@@ -1,13 +1,87 @@
...
+		pageSize, page := 0, 0
+		if params.PageSize != nil {
+			pageSize = *params.PageSize
+		}
+		if params.Page != nil {
+			page = *params.Page
+		}
+
+		result, err := application.ListAnnouncements(r.Context(), r.PathValue("namespace"), pageSize, page)
+		if err != nil {
```
antonve: we probably need some helpers to avoid a lot of duplicated code in the http layer

services/tadoku-api/transport/http/BUILD.bazel
```diff
@@ -16,6 +16,7 @@ go_library(
         "//services/tadoku-api/generated/openapi",
         "//services/tadoku-api/internal/identity",
         "//services/tadoku-api/internal/permissions",
+        "@com_github_deepmap_oapi_codegen//pkg/runtime",
```
antonve: is this the right package to use? shouldn't the generated openapi code be enough?

services/tadoku-api/e2e/routing_test.go
```diff
@@ -111,13 +112,34 @@ func TestReadFailureDoesNotFallBack(t *testing.T) {
 	})
 	api.db.Pool.Close()
```
antonve: what are we exactly testing in routing test? it's becoming less clear to me (tbh i haven't looked at this in detail but this test looks very sus to me)

services/tadoku-api/e2e/legacy_content_test.go
```diff
@@ -75,8 +76,13 @@ type legacyClockConn struct {
 }
 
 func (c *legacyClockConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
```
antonve: what are we doing here??

services/tadoku-api/e2e/announcements_test.go
```diff
@@ -1,14 +1,120 @@
...
+					if api.proxied.Load() != 0 {
+						t.Error("native read contacted an upstream")
+					}
+				})
+			}
+		})
+	}
+}
+
+func TestListAnnouncementsAuthorizationWiring(t *testing.T) {
```
antonve: i don't like how these tests are not standard e2e tests like the above

services/tadoku-api/transport/http/BUILD.bazel
antonve: Yes—after rebasing onto merged #874, the admin list now implements the generated strict server interface. The handwritten query binder and direct legacy runtime dependency in this HTTP package are gone; generated code owns routing, parameter decoding, and response serialization.

The existing old/new goldens pass unchanged, including malformed and negative pagination. All 38 backend test targets passed, plus ten shuffled race-enabled repetitions of the six focused targets.

Updated head: [01880681](https://github.com/tadoku/tadoku/commit/01880681a9b480f5d8bb935caf28d7149472c931).

**Posted by gpt-6-astra**

services/tadoku-api/cmd/tadoku-api/main.go
```diff
@@ -126,9 +127,13 @@ func start(cfg config, logger *slog.Logger) (*application, error) {
 		}
 	}()
 
+	keto := ketoclient.NewReadClient(cfg.KetoReadURL)
+	permissionChecker := permissions.NewChecker(func(ctx context.Context, subjectID string) (bool, error) {
```
antonve: can't we export this function from permissions package itself injecting the keto client?

services/tadoku-api/app/announcements.go
antonve: Yes. The callback can use a named `err` result; it does not need a separate `var err`. I’ll apply that small change while keeping the transaction result assignment unchanged.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/announcements.go
antonve: The shared HTTP plumbing is now handled by the generated server and the shared response-error handler: routing, parameter decoding, response serialization, and permission-error status mapping are no longer repeated per endpoint. The remaining announcement-to-DTO mapping is explicit; I’m not introducing a generic mapper or changing the layer boundaries in this PR.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/routing_test.go
antonve: This test grew to cover several different things: authorization before database access, database failure without proxy fallback, and probe behavior. That makes its purpose unclear. I’ll separate the isolated dependency-failure checks from the routing tests, use explicit test names, and keep normal endpoint cases in the shared HTTP fixture runner.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: Those branches let the admin list’s count/list SQL reach the real database unchanged. The wrapper exists because the legacy active-list query uses PostgreSQL `now()`; only that query has its clock input bound to the scenario time. The query-name/argument checks guard that test-only clock substitution. I’ll narrow the interception to the clock-dependent query so ordinary queries don’t need a growing allowlist; its clock-shape guard will remain.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/announcements_test.go
antonve: Agreed. These cases currently rewrite requests from the permission-test fixtures and check only the status. I’ll replace that with endpoint-specific `request.http`, `golden.http`, and Keto relationship fixtures, using the standard naming/reset/runner flow. The failed-ban-lookup case will also compare a full golden response; its synthetic failure context will stay explicit.

**Posted by gpt-6-astra**

services/tadoku-api/cmd/tadoku-api/main.go
antonve: Yes. I’ll add a Keto-backed constructor in `permissions` that accepts the existing client, and use it from startup and the real-Keto E2E setup. The permission tuple mapping will live there instead of being repeated in composition code. The lower-level function injection can remain for isolated permission checks.

**Posted by gpt-6-astra**

services/tadoku-api/app/announcements.go
antonve: Done in [9c6365e1](https://github.com/tadoku/tadoku/commit/9c6365e1490bdd0fd1c6386487f69812a7a2b8ca): the transaction callback now has a named `err` result, with no separate declaration. The full backend suite and repeated race-enabled tests pass.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/routing_test.go
antonve: Done in [07c244b2](https://github.com/tadoku/tadoku/commit/07c244b2402a002c208b8fd93f329d7bb72494b6). `routing_test.go` now covers route registration, method handling, and proxy routing. The closed-database checks moved to `dependency_failure_test.go`, with named subtests for permission denial before database access, read failures without proxy fallback, and liveness/readiness behavior. Cancellation is a separately named test there too.

Endpoint failures now compare full HTTP goldens. Only the destructive dependency test creates an isolated API; normal scenarios keep the suite-level router/database. All 38 backend test targets and ten shuffled, race-enabled repetitions of the six focused targets pass.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: Done in [07c244b2](https://github.com/tadoku/tadoku/commit/07c244b2402a002c208b8fd93f329d7bb72494b6): the two admin-query allowlist branches are gone. Ordinary queries go straight to PostgreSQL unchanged; only `ListActiveAnnouncements` is intercepted to bind its clock input. That substitution still rejects changed clock/argument shapes.

The tests cover ordinary-query passthrough, scenario-clock changes on the same connection, and rejection of an unexpected clock-query shape. Existing old/new endpoint goldens pass unchanged.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/announcements_test.go
antonve: Done in [07c244b2](https://github.com/tadoku/tadoku/commit/07c244b2402a002c208b8fd93f329d7bb72494b6). The admin/non-admin cases now have their own endpoint `request.http` and `golden.http`, with admin relationships seeded alongside them. They use `APITestName`, `resetCase`, and the existing HTTP golden runner through the real JWT/Keto wiring. The request-rewriting helper was removed.

The failed-ban-lookup regression also uses its own full-response golden; only its failure context is injected explicitly. The eight existing old/new contract cases remain unchanged. The E2E suite passes normally and for ten shuffled race-enabled repetitions.

**Posted by gpt-6-astra**

services/tadoku-api/cmd/tadoku-api/main.go
antonve: Done in [7f34504f](https://github.com/tadoku/tadoku/commit/7f34504f187e4924c305812cec5eb63f190933e3): startup and the real-Keto E2E setup now call `permissions.NewKetoChecker(client)`. The constructor owns the admin tuple mapping and accepts a narrow `KetoReader` interface containing only `CheckPermission`.

The lower-level `NewChecker` remains available for isolated capability tests. Tests check tuple/context forwarding and fail-closed errors, and the real-Keto E2Es pass. Depolicy allows the exact existing Keto-client dependency for `permissions`, without opening it to all internal packages; 50 Go files checked, zero violations.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: this still has query specific overrides - if there is an issue in the original code we should probably inject a clock there.

services/tadoku-api/e2e/legacy_content_test.go
```diff
@@ -101,17 +104,22 @@ func TestLegacyClockFollowsScenarioTime(t *testing.T) {
 	})
 }
 
-func TestLegacyClockRejectsUnexpectedQueries(t *testing.T) {
+func TestLegacyClockPassesOtherQueriesThrough(t *testing.T) {
```
antonve: we shouldn't need to have these useless tests on the old content api

services/tadoku-api/e2e/legacy_content_test.go
antonve: Agreed. I’ll remove the query-specific connector entirely and inject the existing `commondomain.Clock` into the legacy active-list service. It will pass one time value into the normal SQLC query, with production using its existing UTC clock and E2Es using the scoped test clock. This keeps the query itself real and removes runtime SQL rewriting from parity tests.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: I’ll delete the legacy clock-wrapper tests along with that wrapper. The shared HTTP goldens will continue exercising both production implementations; I won’t keep a separate suite testing an adapter that no longer exists.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: Done in [3dba5f77](https://github.com/tadoku/tadoku/commit/3dba5f7786ec278e95e2ceee9a605e6c0dc8e9e0). The SQL-rewriting connector is deleted. The legacy active-list service now receives the existing `commondomain.Clock`, reads it once per operation and passes that value into the normal SQLC query. Production uses its existing UTC application clock; parity tests supply the same scoped clock used by Tadoku API. This explicitly changes the legacy time source from database `now()` to its injected application clock, without changing the filter boundaries.

Regenerated SQLC output is included. Existing response goldens pass unchanged, now with real JWT/Keto auth in both APIs. All 38 backend test targets pass, plus ten shuffled race-enabled repetitions of seven focused targets.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/legacy_content_test.go
antonve: Done in [3dba5f77](https://github.com/tadoku/tadoku/commit/3dba5f7786ec278e95e2ceee9a605e6c0dc8e9e0): all three legacy clock-wrapper tests and the wrapper itself are removed. The legacy setup now only assembles the real database, injected clock, auth stack and generated routes. The shared HTTP goldens still test both APIs; the existing legacy domain tests were only adjusted for the constructor/query-input change.

The full backend suite and repeated race-enabled checks pass.

**Posted by gpt-6-astra**

antonve:
Fixed the reproduced intermittent E2E failure in 8c4aea7e.

The real Keto fixture could exceed its shutdown deadline while flooding its log. Its log and listener files shared the watched configuration directory; the [pinned Keto watcher](https://github.com/ory/keto/blob/v25.4.0/oryx/watcherx/file.go) logs directory events, creating a feedback loop. Config now lives in its own directory. A regression check failed before this fix and passes afterward.

No authentication bypasses, golden changes, or increased timeouts. fc4b9821 also makes CI print failed Bazel test output.

Verified:
- All 38 backend test targets pass.
- All eight CI race-test targets pass with ten repetitions and the previously failing shuffle seed.
- Gazelle and formatting checks pass.
- [PR CI](https://github.com/tadoku/tadoku/actions/runs/35084546448) and [push CI](https://github.com/tadoku/tadoku/actions/runs/35084541266) are green on 8c4aea7e; SQLC and docs checks also pass.

**Posted by gpt-6-astra**

PR #874, opened 2026-09-15
Generate Tadoku API strict HTTP server bindings

antonve:
Server-generation infrastructure first, before finishing #873. This PR starts from main; #873 is unchanged and should be adapted after this merges.

## Change

- Generate OpenAPI component models, standard-library HTTP routing, parameter binding, strict request types, and response serialization with one pinned v2 invocation and one output file.
- Register only application-owned operations through the existing Router, preserving shared deadlines, authentication, and ban checks. Other operations and unclaimed methods remain on the legacy proxy.
- Preserve exact HTTP golden responses with the generated-handler-only JSON charset writer, existing error response shapes, and explicit nullable-field mapping.
- Remove the v1 generation command and the four legacy services' generator entrypoints. Their checked-in generated handlers remain frozen until retirement.
- Keep the canonical specification, app/features/repositories, SQL, and all existing HTTP fixtures unchanged. No admin-list implementation or new business behavior.

## Generator isolation

The separate Bazel tool module pins oapi-codegen v2.8.0 for both models and server generation. Application dependencies use runtime v1.6.0 and its official nullable v1.1.0 type. Nullable mapping preserves explicit null, empty-string, and populated responses without changing the specification or goldens.

The root v1 module and legacy kin-openapi pin remain only because the old services still import their runtime/types packages. Removing that remaining dependency after legacy retirement is recorded in the migration log. Root REPO.bazel excludes the nested tool module from root package traversal, and CI watches the boundary and tool-module inputs.

## Verification and review

Implemented by gpt-5.6-sol with high reasoning; independently reviewed by gpt-6-astra with high reasoning. The latest revision has no outstanding review findings.

- Full backend build: 212 targets passed.
- Full backend tests: all 38 test targets passed, including real PostgreSQL/Keto E2Es and unchanged old/new HTTP goldens.
- Independent verification: HTTP E2E, transport, permissions, and startup targets passed ten shuffled repetitions with the race detector.
- Generated nullable-field serialization tests cover null, empty string, and populated values.
- OpenAPI regeneration was byte-stable; all 76 canonical component schemas remain generated.
- Root and isolated generator frozen dependency checks passed.
- Gazelle passed; root package traversal excludes the nested tool module.
- Depolicy: zero violations.
- Shell syntax, formatting, and diff checks passed.

The generator emits a non-fatal Go-version discovery warning from Bazel's runfiles directory; both toolchains are explicitly pinned to Go 1.26.6 and the generated code builds and tests successfully.

No deployment or merge performed.

**Posted by gpt-6-astra**


scripts/generate-openapi.sh
```diff
@@ -3,8 +3,19 @@ set -euo pipefail
 
 API_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "$API_ROOT"
-# The existing generator version is pinned in go.mod. Bazel owns its toolchain.
+# The legacy DTO generator version is pinned in the root go.mod. Bazel owns its
+# toolchain, and its output remains available to every retained operation.
 bazel run @com_github_deepmap_oapi_codegen//cmd/oapi-codegen -- \
```
antonve: we are not gonna use the legacy oapi codegen tool anymore - i believe we can delete it (all the old services will be deleted post migration)

services/tadoku-api/spec/server-codegen.yaml
```diff
@@ -0,0 +1,8 @@
+package: openapi
+generate:
+  models: false
```
antonve: what kind of models are we talking about here?

antonve: “Models” means the Go HTTP contract types generated from the OpenAPI schemas: request/response structs, enums, and parameter types—not database models or SQLC rows.

It was false here because the separate v1 invocation generated those types. I’m removing that split and enabling v2 model generation alongside the server bindings. Component schema types will be retained without registering the still-proxied operations.

**Posted by gpt-6-astra**

scripts/generate-openapi.sh
antonve: Removed in [9c62a91f](https://github.com/tadoku/tadoku/commit/9c62a91f34346368d52de17c3d60561bbcfe0603). The script now runs only v2, generating models and server bindings together. I also removed the four legacy services' generator entrypoints; their checked-in generated handlers stay frozen.

The v1 module still supplies runtime/types packages imported by those legacy handlers, so that dependency must remain until the services are retired. Its removal is recorded in the migration log.

All 38 backend test targets passed, including unchanged old/new goldens; independent race/shuffle verification also passed.

**Posted by gpt-6-astra**

services/tadoku-api/spec/server-codegen.yaml
antonve: Implemented in [9c62a91f](https://github.com/tadoku/tadoku/commit/9c62a91f34346368d52de17c3d60561bbcfe0603): `models: true`, with models and server bindings generated together by v2. All component-schema types are retained; only migrated operations receive server bindings.

The supported nullable option preserves the existing null/empty-string/value responses. The spec and golden fixtures are unchanged.

**Posted by gpt-6-astra**

scripts/generate-openapi.sh
```diff
@@ -2,9 +2,10 @@
...
 
 API_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
-cd "$API_ROOT"
-# The existing generator version is pinned in go.mod. Bazel owns its toolchain.
-bazel run @com_github_deepmap_oapi_codegen//cmd/oapi-codegen -- \
-  -generate types,skip-prune -package openapi \
-  -o "$API_ROOT/services/tadoku-api/generated/openapi/api.gen.go" \
-  "$API_ROOT/services/tadoku-api/spec/openapi.yaml"
+(
+  cd "$API_ROOT/tools/oapi-codegen-v2"
```
antonve: why are we still calling it v2? i want the new version to replace the old one

antonve: That name was left over from the initial split. Renamed it to `tools/oapi-codegen` in [6a2de083](https://github.com/tadoku/tadoku/commit/6a2de0833bd8a8aece64009426348cec973574a1), including the local module/SDK names, CI paths, and documentation.

This is the default and only generator. Version suffixes remain only in the upstream dependency identity/version, not our tooling name.

Regeneration produces identical output; frozen dependency checks, Gazelle, and the root build exclusion pass.

**Posted by gpt-6-astra**

PR #878, opened 2026-09-16
feat(tadoku-api): migrate announcement lookup by ID

antonve:
## Summary

Migrate `GET /content/announcements/{namespace}/{id}` into Tadoku API using the generated strict server and existing Content feature.

- Require admin permission in the application operation; keep shared JWT and ban middleware unchanged.
- Query by namespace and ID, excluding soft-deleted records. Preserve the legacy response mapping, empty malformed-ID 400 and missing-record 404. Use a shared invalid-UUID error mapped to 400 by the generic response-error handler.
- Keep HEAD on the legacy proxy using a proxy-only dispatcher, avoiding conflicting wildcard HEAD/GET registrations.
- Rebased onto the new suite/fixture ownership structure. Eight shared golden cases run against both APIs with real JWT, Keto and PostgreSQL; additional checks cover failed storage, permission-before-storage, transaction context and proxy ownership.
- Regenerate OpenAPI, SQLC and Bazel metadata. No migrations, new dependencies, write endpoints, or deferred database-helper changes.

Flow: generated handler → application admin check → Content service → AnnouncementsRepository → SQLC. This is one read query, so it does not open a new transaction.

## Verification

- `bazel test //services/... --test_output=errors` — all 38 targets passed.
- Eight CI race-test targets with shuffle enabled and count 5 — passed.
- `bazel run //:gazelle -- -mode=diff` — passed.
- `bazel run //tools/ci/depolicy` — zero violations.
- `pnpm --dir docs api:check` — passed for the production change; no contract/docs changes in the test update.
- SQLC and OpenAPI regeneration completed with the checked-in outputs.

No production deployment.

**Posted by gpt-6-astra**


services/tadoku-api/transport/http/router.go
```diff
@@ -92,6 +94,8 @@ func NewHandler(
 			ResponseErrorHandlerFunc: func(w stdhttp.ResponseWriter, _ *stdhttp.Request, err error) {
 				status := stdhttp.StatusInternalServerError
 				switch {
+				case errors.Is(err, errInvalidAnnouncementID):
```
antonve: this sets a bad precedent: we should not add specific errors here. Write the status code directly in the original handler instead. this endpoint is for shared errors and generic errors.

antonve: Fixed in 1947da14. The announcement handler now returns its own empty-400 response, with the status written locally in announcements.go. Removed the endpoint-specific error sentinel and its case from the shared response-error handler; that handler keeps only shared permission errors and the generic fallback.

The existing malformed-ID golden still passes against both APIs. E2E and transport tests pass; Depolicy reports zero violations.

**Posted by gpt-6-astra**

antonve: Updated in e5344d9a per the latest discussion: handlers now return a shared `errInvalidUUID`, and the generic response-error handler maps it to an empty HTTP 400 via `errors.Is`. The announcement-specific response type is removed; no announcement-specific errors were added to the router.

The stacked delete and update endpoints use the same error. Existing old/new API goldens are unchanged. #878's E2E and transport tests pass; all 38 backend test targets pass on the combined stack, with Gazelle and Depolicy clean.

**Posted by gpt-6-astra**

PR #879, opened 2026-09-16
feat(tadoku-api): migrate announcement deletion

antonve:
## Summary

Targets `main`; the lookup support from #878 is already merged.

Migrate `DELETE /content/announcements/{namespace}/{id}` through the generated strict handler, application admin check, Content service and AnnouncementsRepository.

- Preserve legacy idempotent 204 responses for missing, already-deleted and wrong-namespace rows.
- Soft-delete using business time; reuse the shared invalid-UUID error and generic HTTP 400 mapping from #878.
- Add shared old/new API goldens with real JWT, Keto and PostgreSQL, plus a focused real-PostgreSQL repository test for soft deletion, row preservation, namespace/ID filtering and repeated-delete timestamps.
- Keep unclaimed methods proxied. No migrations or new dependencies.

## Verification

- `bazel test //services/... --test_output=errors` — all 38 targets passed.
- OpenAPI and SQLC generation — complete output committed.
- Content and E2E race tests with shuffle and count 5 — passed.

**Posted by gpt-6-astra**


services/tadoku-api/e2e/announcement_delete_test.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func TestDeleteAnnouncement(t *testing.T) {
+	tests := []struct {
+		description []string
+		want        int
+	}{
+		{description: []string{"admin"}, want: http.StatusNoContent},
+		{description: []string{"missing"}, want: http.StatusNoContent},
+		{description: []string{"already", "deleted"}, want: http.StatusNoContent},
```
antonve, line 15: shouldn't delete fail when there is nothing to delete?

services/tadoku-api/e2e/announcement_delete_test.go
```diff
@@ -0,0 +1,94 @@
...
+		t.Run(name, func(t *testing.T) {
+			runCase(t, api, name, test.want,
+				implementation{name: "tadoku-api", handler: api.handler},
+				implementation{name: "content-api", handler: legacyContent.handler},
+			)
+		})
+	}
+}
+
+func TestDeleteAnnouncementPreservesRow(t *testing.T) {
```
antonve: There is no need to test this. This should be a repository test (do we have those?)

services/tadoku-api/e2e/dependency_failure_test.go
```diff
@@ -32,6 +32,20 @@ func TestDependencyFailures(t *testing.T) {
 		suite       *suite
 		handler     http.Handler
 	}{
+		{
+			operation:   "DeleteAnnouncement",
+			description: []string{"non", "admin"},
+			want:        http.StatusForbidden,
+			suite:       poolClosed,
```
antonve: Do we really need more of these? We already tested this on a couple of endpoints, we don't really need like a whole test suite testing every dependency going away for every endpoint.

services/tadoku-api/features/content/announcement_delete_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/tadoku/tadoku/services/tadoku-api/features/content"
+	"github.com/tadoku/tadoku/services/tadoku-api/infra/postgres"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/testpostgres"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/timex"
+)
+
+func TestDeleteAnnouncementUsesBusinessTimeAndTransaction(t *testing.T) {
```
antonve: Do we really want to test this? This doesn't seem all that useful to me.

services/tadoku-api/e2e/announcement_delete_test.go
antonve: The legacy API currently returns 204 when no row matches, including missing, already-deleted and wrong-namespace announcements. Its [delete query](https://github.com/tadoku/tadoku/blob/e75571ff7e172155ee876695202d2d47d3a65525/services/content-api/storage/postgres/queries/announcements.sql#L56-L61) uses `:exec`, and the repository never checks the affected-row count. Although the handler has a 404 branch, that no-match path does not produce its not-found error.

I've kept 204 here because this PR is a parity migration. Returning 404 would be an intentional behavior change, not a fix needed to match the old API. Leaving this thread open for your decision on that change.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/announcement_delete_test.go
antonve: Moved this into `TestAnnouncementsRepositoryDeleteAnnouncement` in `features/content/announcements_repository_test.go` (42e77691). The focused real-PostgreSQL test checks that the row survives, only `deleted_at` changes, namespace/ID filtering is respected, and repeated deletion preserves the first timestamp.

Removed the HTTP-level row inspection. The eight shared old/new response goldens remain; no SQL persistence assertions are left in the delete E2E.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/dependency_failure_test.go
antonve: Removed both delete-specific closed-pool cases and their now-unused 500 fixture in 42e77691. This PR no longer adds anything to `TestDependencyFailures`; existing shared-boundary coverage stays in place.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_delete_test.go
antonve: Removed `TestDeleteAnnouncementUsesBusinessTimeAndTransaction` and its file in 42e77691. We no longer repeat clock, rollback or cancellation checks for this operation. The remaining repository test covers the actual delete query; clock and transaction helpers retain their own tests.

All 38 backend test targets pass. The content and E2E targets also pass with the race detector, shuffled order and five repetitions.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/announcement_delete_test.go
antonve: Ack, leave a note in our post migration todo to fix this behavior (and probably also the other content endpoints)

PR #880, opened 2026-09-17
feat(tadoku-api): migrate announcement creation

antonve:
## Summary

Targets `main`; #878 and #879 are already merged.

Migrate `POST /content/announcements/{namespace}` through the generated strict handler, application admin check, Content service and AnnouncementsRepository.

- Use generated JSON decoding only, with no legacy body adapter or non-JSON request fixtures.
- Validate `CreateAnnouncementParameters` with shared feature-domain style validation and `timex.IsValidRange`.
- Use named `internal/errx` constructors carrying `Kind`, message and cause, such as `NewInvalidInputError(message)` and `NewUnavailableError(message, cause)`. The HTTP boundary maps metadata to status; unknown failures remain 500. Legacy error handling is unchanged.
- Preserve nullable fields, optional UUID generation and business-time timestamps. The legacy INSERT persists timestamps already supplied by its domain layer.
- Keep creation HTTP cases in one real JWT/Keto/PostgreSQL golden table. Direct repository tests cover persistence; pure validation stays in unit tests. No extra generated-ID/readback HTTP tests or repeated dependency-failure matrix.
- Empty JSON requests from signed guest/non-admin identities return 400 before application authorization. Those two intentional legacy differences are explicit in the parity table.
- No new external dependencies or schema migrations.

## Verification

- Latest constructor follow-up: `bazel test //services/tadoku-api/... --test_output=errors` — all 11 targets passed against disposable local PostgreSQL.
- Earlier typed-error round: `bazel test //services/... --test_output=errors` — all 39 targets passed.
- Errx, permissions, transport and HTTP E2E targets with `--@rules_go//go/config:race --test_arg=-test.shuffle=on --test_arg=-test.count=5` — all four passed.
- No SQL, OpenAPI or golden inputs changed in this review round.

**Posted by gpt-6-astra**


services/tadoku-api/e2e/dependency_failure_test.go
```diff
@@ -32,6 +32,20 @@ func TestDependencyFailures(t *testing.T) {
 		suite       *suite
 		handler     http.Handler
 	}{
+		{
```
antonve: Stop writing these tests if the same dependencies have been tested before - make this a comment above this test so you don't repeat it

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,53 @@
...
+	"errors"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/timex"
+)
+
+var ErrInvalidAnnouncement = errors.New("invalid announcement")
+
+type CreateAnnouncementRequest struct {
```
antonve: Internally within the service we shouldnt call it a request, how about CreateAnnouncementParameters?

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,53 @@
...
+	StartsAt  time.Time
+	EndsAt    time.Time
+}
+
+func (s *Service) CreateAnnouncement(ctx context.Context, request CreateAnnouncementRequest) (*Announcement, error) {
+	if request.ID == uuid.Nil || request.Namespace == "" || request.Title == "" || request.Content == "" ||
+		request.StartsAt.IsZero() || request.EndsAt.IsZero() || !request.EndsAt.After(request.StartsAt) {
+		return nil, ErrInvalidAnnouncement
+	}
+	switch request.Style {
```
antonve: I want this validation for style to be a function in the feature domain itself, edit will also need this

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,53 @@
...
+	Content   string
+	Style     string
+	Href      *string
+	StartsAt  time.Time
+	EndsAt    time.Time
+}
+
+func (s *Service) CreateAnnouncement(ctx context.Context, request CreateAnnouncementRequest) (*Announcement, error) {
+	if request.ID == uuid.Nil || request.Namespace == "" || request.Title == "" || request.Content == "" ||
+		request.StartsAt.IsZero() || request.EndsAt.IsZero() || !request.EndsAt.After(request.StartsAt) {
```
antonve: Can we have a reusable function in a datex package shared with everyone? (Like timex) i want a reusable function to validate time ranges

services/tadoku-api/features/content/announcement_create_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/tadoku/tadoku/services/tadoku-api/features/content"
+	"github.com/tadoku/tadoku/services/tadoku-api/infra/postgres"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/testpostgres"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/timex"
+)
+
+func TestCreateAnnouncementValidationBeforeStorage(t *testing.T) {
```
antonve: Just call it TestCreateAnnouncementParametersValidation() and extract the validation as a function on the struct?

services/tadoku-api/features/content/announcement_create_test.go
```diff
@@ -0,0 +1,136 @@
...
+			test.change(&request)
+			_, err := content.NewService(nil).CreateAnnouncement(t.Context(), request)
+			if !errors.Is(err, content.ErrInvalidAnnouncement) {
+				t.Errorf("error=%v, want invalid announcement", err)
+			}
+		})
+	}
+}
+
+func TestCreateAnnouncementBusinessTimeAndRepositoryTransactions(t *testing.T) {
```
antonve: I dont sant tests with a database at the feature service level - i would delete and rely on the e2e tests

services/tadoku-api/features/content/announcements_repository.go
```diff
@@ -42,6 +42,34 @@ func (r *AnnouncementsRepository) DeleteAnnouncement(ctx context.Context, namesp
 	return nil
 }
 
+func (r *AnnouncementsRepository) CreateAnnouncement(ctx context.Context, item *Announcement) error {
```
antonve, line 45: Write repository test, also document that repository tests are highly recommended, prefer repository tests and e2e test combination over testing feature service in isolation.

services/tadoku-api/transport/http/announcement_body.go
```diff
@@ -0,0 +1,74 @@
...
+	"encoding/json"
+	"encoding/xml"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/tadoku/tadoku/services/tadoku-api/generated/openapi"
+)
+
+func isAnnouncementWriteRequest(r *http.Request) bool {
```
antonve: Wtf are we doing??

services/tadoku-api/transport/http/announcement_create.go
```diff
@@ -0,0 +1,64 @@
...
+		ID:        id,
+		Namespace: request.Namespace,
+		Title:     body.Title,
+		Content:   body.Content,
+		Style:     string(body.Style),
+		Href:      href,
+		StartsAt:  body.StartsAt,
+		EndsAt:    body.EndsAt,
+	})
+	if errors.Is(err, app.ErrInvalidAnnouncement) {
```
antonve: I wonder if certain errors could be decorated with some kind of annotation or something to indicate status? how is this often done?

antonve (commented):
Fix these

services/tadoku-api/e2e/dependency_failure_test.go
antonve: Removed the creation-specific closed-pool cases and their unused fixture. Added the requested comment above `TestDependencyFailures`: cover a shared dependency failure once, not again for each endpoint. The stacked update PR (#881) also no longer repeats those outage cases. Shared boundary coverage remains.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: Renamed it to `CreateAnnouncementParameters` in the feature, app alias and handler construction. Transport-generated request objects keep their transport names. Applied the matching `UpdateAnnouncementParameters` naming in #881.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: Extracted `content.IsValidAnnouncementStyle`, used by `CreateAnnouncementParameters.Validate`. #881 now reuses it for update validation too.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: Added `internal/datex.IsValidRange(start, end)`, available throughout tadoku-api like `timex`. It requires both endpoints to be set and `start` to precede `end`, preserving the existing rule. It has pure tests for missing/equal/reversed endpoints, valid ranges and equivalent instants in different time zones. Creation and #881's update both use it.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create_test.go
antonve: Validation is now `CreateAnnouncementParameters.Validate()`, and the pure table test is named `TestCreateAnnouncementParametersValidation`. It calls validation directly—no service or database setup.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create_test.go
antonve: Deleted the database-backed creation feature-service test. The HTTP E2Es retain the real application path and committed readback checks; pure parameter validation has its own unit test. Database-specific persistence assertions now call the repository directly.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcements_repository.go
antonve: Added `TestAnnouncementsRepositoryCreateAnnouncement` against real PostgreSQL. It directly checks persisted fields/timestamps, null/empty/populated hrefs, duplicate-ID rejection and preservation of the original row. Documented in AGENTS.md and the API README that repository tests plus HTTP E2Es are preferred, with no database-backed feature-service tests. All 39 backend test targets and the focused race/shuffle runs passed.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/announcement_body.go
antonve: This was adapting the legacy binder's XML/form and empty-body behavior before the generated JSON decoder. Per your clarification in chat, it is now deleted entirely. The new API uses its generated JSON decoder; the non-JSON fixtures are removed, with no replacement non-JSON test matrix. Empty JSON requests from signed guest/non-admin identities now return 400 rather than the legacy 401/403; those intentional differences are explicit in the parity table. Applied the same cleanup to #881.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/announcement_create.go
antonve: Yes—an error can carry metadata in a custom type, expose the original cause with `Unwrap()`, and be inspected using `errors.As`. That is the normal Go mechanism for this kind of annotation. [Go's error-wrapping guide](https://go.dev/blog/go1.13-errors) explains the pattern.

Two options fit here:

- A transport-neutral category such as invalid input/not found, mapped to an HTTP status once at the HTTP boundary.
- An HTTP-specific wrapper carrying a status, created only in the HTTP layer.

I prefer the first for feature/app errors: it avoids coupling those layers to HTTP. We already have a similar central mapping in `services/common/http/httperr.StatusCode`; this handler currently maps its feature sentinel into the generated typed 400 response explicitly.

I have not added a new error taxonomy or wrapper in this PR. Leaving this discussion open so we can agree on that design separately; the concrete review fixes are pushed.

**Posted by gpt-6-astra**

services/tadoku-api/e2e/announcement_create_test.go
```diff
@@ -0,0 +1,144 @@
...
+			}
+			runCase(t, api, name, test.want,
+				implementation{name: "tadoku-api", handler: api.handler},
+				legacy,
+			)
+		})
+	}
+}
+
+func TestCreateAnnouncementCommitsForReadback(t *testing.T) {
```
antonve: We dont need the tests below this - I want to be disciplined about the kind of tests we introduce. I want them to be focused in a single table with all test cases.

Seriously- stop adding useless tests, we are trying to set a golden example of how to implement features in this repo. We need to be disciplined or else agents will copy bad practices

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,56 @@
...
+	Content   string
+	Style     string
+	Href      *string
+	StartsAt  time.Time
+	EndsAt    time.Time
+}
+
+func (p CreateAnnouncementParameters) Validate() error {
+	if p.ID == uuid.Nil || p.Namespace == "" || p.Title == "" || p.Content == "" ||
+		!IsValidAnnouncementStyle(p.Style) || !datex.IsValidRange(p.StartsAt, p.EndsAt) {
```
antonve: Actually i just realized datex functions can be moved to timex, nearly same thing anyway

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,56 @@
...
+		Content:   parameters.Content,
+		Style:     parameters.Style,
+		Href:      parameters.Href,
+		StartsAt:  parameters.StartsAt,
+		EndsAt:    parameters.EndsAt,
+		CreatedAt: now,
+		UpdatedAt: now,
+	}
+	if err := s.announcements.CreateAnnouncement(ctx, item); err != nil {
+		return nil, err
```
antonve, line 52: Q: do we need error wrapping?

services/tadoku-api/transport/http/announcement_create.go
antonve: Let's go with your first suggestion

services/tadoku-api/e2e/announcement_create_test.go
antonve: Removed both extra creation HTTP tests and their now-unused fixtures. `TestCreateAnnouncement` is the sole creation E2E table; persistence assertions stay in the direct repository tests. Added this as a general convention in AGENTS.md and the API README, and removed the separate update readback tests from #881 as well. No replacement bespoke test harness or non-JSON cases.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: Moved the helper and its existing tests into `internal/timex` as `timex.IsValidRange`. Removed the `datex` package and its build/policy entries. Creation and the stacked update PR (#881) both use the moved helper.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/announcement_create.go
antonve: Implemented the transport-neutral category approach using the existing `services/common/domain` errors and `services/common/http/httperr.StatusCode`—no new error framework.

Feature-specific sentinels wrap invalid-input/not-found categories with `%w`, so `errors.Is` still recognizes the specific error and its category. Permission errors use the same shared categories. The HTTP router maps categories once, with unclassified failures returning 500; per-handler feature-error status switches and their app aliases are removed. #881 uses the same path.

Existing JSON goldens remain unchanged. All 38 backend test targets, five focused race targets with shuffle/count 5, Gazelle and Depolicy passed on both branches.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: Not at this return site. The HTTP handler already logs which operation failed, and returning the repository error unchanged preserves its cause. Adding another generic “create announcement failed” wrapper here would duplicate that context.

The new `%w` wrapping on the feature sentinels serves a different purpose: it attaches the approved transport-neutral category. We should wrap call-site errors only when adding useful context or deliberately classifying them, not at every layer.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+	commondomain "github.com/tadoku/tadoku/services/common/domain"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/timex"
+)
+
+var ErrInvalidAnnouncement = fmt.Errorf("invalid announcement: %w", commondomain.ErrRequestInvalid)
```
antonve: Is this really how you do this? i remember seeing something about attaching metadata to errors that are later read by the error mapping code...

services/tadoku-api/internal/permissions/checker.go
```diff
@@ -3,17 +3,17 @@ package permissions
...
 	ketoclient "github.com/tadoku/tadoku/services/common/client/keto"
+	commondomain "github.com/tadoku/tadoku/services/common/domain"
 	"github.com/tadoku/tadoku/services/tadoku-api/internal/identity"
 )
 
 var (
-	ErrUnauthorized = errors.New("unauthorized")
-	ErrForbidden    = errors.New("forbidden")
-	ErrUnavailable  = errors.New("permissions unavailable")
+	ErrUnauthorized = commondomain.ErrUnauthorized
```
antonve: Why not just return the common domain errors instead of reassigning them here?

services/tadoku-api/features/content/announcement_create.go
antonve: Yes—you’re describing a typed error carrying metadata. The current implementation uses **sentinel classification**, not a metadata field: `%w` preserves the category error, and the mapper recognizes it with `errors.Is`. It does not inspect the error message.

The metadata version would use a small error type with a transport-neutral `Kind` (for example, invalid input or not found), a message and an optional cause. The mapper would retrieve it with `errors.As` and map `Kind` to an HTTP status; `Unwrap` would preserve the cause. Both approaches are supported by Go’s [error inspection and wrapping mechanisms](https://go.dev/blog/go1.13-errors).

I reused the existing sentinels to avoid adding a second error contract, but that is not the explicit metadata representation you were recalling. Do you want me to switch this PR to the typed `Kind` approach? I’ve left this thread open rather than treating the explanation as approval to change it.

**Posted by gpt-6-astra**

services/tadoku-api/internal/permissions/checker.go
antonve: Agreed—the re-exports added no value. Removed all three permission-local errors. The checker now returns `commondomain.ErrUnauthorized`, `ErrForbidden` and `ErrAuthzUnavailable` directly; provider failures still wrap the common category and original cause with the lookup context. Updated the existing assertions rather than adding tests.

Pushed in `fd828200`, and restacked #881. All 10 Tadoku API test targets, Gazelle consistency and Depolicy passed on both branches; HTTP goldens are unchanged.

**Posted by gpt-6-astra**

services/tadoku-api/features/content/announcement_create.go
antonve: I want to change the approach in tadoku-api to use the Kind approach

antonve: Implemented in `0d8e0931`, using `internal/errx` as agreed in chat.

- `errx.Error` carries `Kind`, `Message` and optional `Cause`.
- `errx.KindOf` reads the typed metadata through `errors.As`; HTTP status mapping stays in the transport layer. Unknown/unclassified errors remain 500.
- Content and permission errors now use `errx`, without the common-domain category sentinels. Causes remain inspectable through `Unwrap`, and an outer classification takes precedence over its cause's kind.
- Legacy error packages and services are unchanged by this refactor; no HTTP goldens changed.

All 39 backend test targets and four focused race targets passed. Restacked #881 and verified all 11 Tadoku API targets there too. Gazelle and Depolicy are clean.

**Posted by gpt-6-astra**

PR #881, opened 2026-09-17
feat(tadoku-api): migrate announcement updates

antonve:
## Summary

Targets `main`; #878, #879 and #880 are merged.

Migrate `PUT /content/announcements/{namespace}/{id}` through the generated strict handler, application admin check and Content feature.

- Carry the URL `ID` and `Namespace` alongside body fields in `UpdateAnnouncementParameters`; app and feature methods take only `(ctx, parameters)`. Validate only in the feature service. The app owns authorization and the read/update transaction. Reuse feature-domain style validation and `timex.IsValidRange`; repositories share the transaction context and only query/map persisted data.
- Reuse #880's named `internal/errx` constructors and centralized kind-to-HTTP-status mapping; legacy error handling is unchanged.
- Use generated JSON decoding only, with no legacy body adapter or non-JSON request fixtures. Nullable href decoding follows the same `Get()` pattern as creation.
- Preserve namespace and soft-delete filtering, nullable fields, ignored server fields and JSON response contracts. Empty JSON requests from signed guest/non-admin identities return 400; these two intentional legacy differences are explicit in the parity table.
- Make the legacy UPDATE persist its existing domain-supplied update timestamp.
- Keep update HTTP cases in one real JWT/Keto/PostgreSQL golden table. Persistence and commit/rollback assertions remain in repository tests; no separate readback HTTP tests, duplicate invalid-UUID handler test or repeated dependency-outage matrix.
- No new external dependencies or schema migrations.

## Verification

- After consolidating update inputs into the parameters struct: `bazel test //services/tadoku-api/... --test_output=errors` — all 11 targets passed against disposable local PostgreSQL.
- The preceding rebased revision passed all 39 backend targets and three focused race/shuffle/count-5 targets.
- No SQL, OpenAPI or golden inputs changed in this review round.

**Posted by gpt-6-astra**


services/tadoku-api/features/content/announcement_update.go
```diff
@@ -0,0 +1,50 @@
...
+type UpdateAnnouncementParameters struct {
+	Title    string
+	Content  string
+	Style    string
+	Href     *string
+	StartsAt time.Time
+	EndsAt   time.Time
+}
+
+func (p UpdateAnnouncementParameters) Validate(namespace string) error {
```
antonve: Why are we not passing in namespace as a parameter?

antonve: Moved both `Namespace` and `ID` into `UpdateAnnouncementParameters` in `e77f8158`, including your follow-up in chat. The HTTP handler fills them from the URL; app and feature methods now take only `(ctx, parameters)`. `Validate()` takes no arguments and is called only by the feature service.

Updated the existing validation test. All 11 Tadoku API test targets passed, including unchanged HTTP goldens; Gazelle and Depolicy are clean.

**Posted by gpt-6-astra**

PR #887, opened 2026-09-17
fix(tadoku-api): store announcement timestamps as UTC instants

antonve:
Compose each announcement write and `FindAnnouncementByID` readback inside one application transaction. Create and update responses use the row read from PostgreSQL, including its stored timestamp precision. Write, readback, or commit failure returns no announcement and propagates the error.

Centralize `time.Time` to `pgtype.Timestamp` conversion in the native PostgreSQL package. The helper normalizes each instant to UTC before writing to the existing `timestamp without time zone` columns and is reused for announcement writes, deletion time and active-list cutoffs.

Use ordinary single-request create/update HTTP goldens with offset-bearing, sub-microsecond timestamps to verify persisted response values. Keep the real-database repository round-trip regression. Offset and precision differences from legacy are documented explicitly in the golden tables.

## Validation

- Ordinary create/update offset cases failed before the fix because responses retained nanoseconds that PostgreSQL does not store; both now pass with the stored microsecond values.
- Focused content repository, PostgreSQL helper/transaction, and HTTP E2E tests pass.
- `bazel test //services/tadoku-api/... --test_output=errors` — all 11 targets pass.
- `bazel run //tools/ci/depolicy` — zero violations.
- `bazel build //services/...` and `bazel test //services/... --test_output=errors` — all 39 test targets pass.
- Gazelle and the complete sqlc generator ran; generated output matches the final query sources.


services/tadoku-api/e2e/announcement_create_test.go
```diff
@@ -51,8 +54,23 @@ func TestCreateAnnouncement(t *testing.T) {
...
+					legacy.skip = "legacy stores timestamp offsets as wall-clock fields"
+				}
 			}
-			runCase(t, api, name, test.want,
+			requests := []goldenRequest{{
+				requestFile: "request.http",
+				goldenFile:  "golden.http",
+				want:        test.want,
+			}}
+			if test.readback {
```
antonve: No i dont want a readback test, i want to change our implementation so that it always does a readback instead of from memory

services/tadoku-api/features/content/announcements_repository.go
```diff
@@ -59,10 +59,10 @@ func (r *AnnouncementsRepository) CreateAnnouncement(ctx context.Context, item *
 		Content:   item.Content,
 		Style:     item.Style,
 		Href:      href,
-		StartsAt:  pgtype.Timestamp{Time: item.StartsAt, Valid: true},
-		EndsAt:    pgtype.Timestamp{Time: item.EndsAt, Valid: true},
-		CreatedAt: pgtype.Timestamp{Time: item.CreatedAt, Valid: true},
-		UpdatedAt: pgtype.Timestamp{Time: item.UpdatedAt, Valid: true},
+		StartsAt:  pgtype.Timestamp{Time: item.StartsAt.UTC(), Valid: true},
```
antonve: Can we make a function to easily convert to timestamp correctly, iirc we have a place for that

services/tadoku-api/app/announcement_create.go
```diff
@@ -13,5 +14,16 @@ func (a *Application) CreateAnnouncement(ctx context.Context, parameters CreateA
...
-	return a.content.CreateAnnouncement(ctx, parameters)
+	var result *content.Announcement
+	err := postgres.RunInTransaction(ctx, a.db, func(ctx context.Context) (err error) {
+		if err := a.content.CreateAnnouncement(ctx, parameters); err != nil {
+			return err
+		}
+		result, err = a.content.FindAnnouncementByID(ctx, parameters.Namespace, parameters.ID)
+		return err
+	})
+	if err != nil {
```
antonve, line 25: Cant we just return result, err here

services/tadoku-api/e2e/announcement_create_test.go
```diff
@@ -51,6 +52,9 @@ func TestCreateAnnouncement(t *testing.T) {
 			legacy := implementation{name: "content-api", handler: legacyContent.handler}
 			if test.skipParity {
 				legacy.skip = "intentional JSON-only decoding difference"
+				if test.want == http.StatusCreated {
```
antonve: We shouldn't identify tests like this

PR #890, opened 2026-09-17
fix(tadoku-api): reject out-of-range pagination

antonve:
Negative announcement pagination values now return 400, while `page_size=0` keeps its default of 10. Pages beyond the result set return an empty list with the true `total_size`, and valid offsets use `int64`/PostgreSQL `bigint` so large result sets are not truncated at `MaxInt32`.

Regression coverage includes negative page inputs, a `MaxInt64` page that previously overflowed before narrowing, and an empty page beyond `total_size`. The sqlc output is regenerated from the updated query.

Validation:

- `bazel test //services/tadoku-api/e2e:e2e_test --test_filter=TestListAnnouncements --test_output=errors`
- `bazel test //services/tadoku-api/... --test_output=errors`


services/tadoku-api/e2e/announcements_test.go
```diff
@@ -46,16 +47,25 @@ func TestListAnnouncements(t *testing.T) {
...
+		{description: []string{"page", "beyond", "total", "size"}, want: http.StatusOK},
 	}
 
 	for _, test := range tests {
 		name := APITestName("ListAnnouncements", test.want, test.description...)
 		t.Run(name, func(t *testing.T) {
+			legacy := implementation{name: "content-api", handler: legacyContent.handler}
+			if test.skipParity {
+				legacy.skip = "legacy overflows pagination offsets"
+				if test.want == http.StatusBadRequest {
```
antonve: do not handle skip reason like this, use a string for skipParity like in 888 i believe

PR #902, opened 2026-09-17
refactor(tadoku-api): move the frozen clock into a test-only package

antonve:
## Summary

- co-locate the frozen-clock helper with `internal/timex` as the test-only `timex_test` package and Bazel target
- make the E2E depend on `//services/tadoku-api/internal/timex:timex_test` and remove the nested `timextest` package
- retain CAS rejection, panic-safe restoration, UTC normalization, concurrent-reader coverage, and the production-only real clock
- restrict the shared clock-state bridge to the containing `internal/timex` Bazel package

The small `clockstate` bridge remains necessary because Go's external `timex_test` package cannot access `timex`'s unexported atomic state. A same-import-path rules_go overlay works for the local unit test but cannot be linked into the E2E, whose transitive production dependencies already contain the normal `timex` archive. The resulting duplicate package is rejected by the linker. Keeping the restricted bridge avoids exported production mutation hooks and unsafe linker mechanisms.

The test helper is `testonly`, visible only to the E2E package, and absent from the production Tadoku API dependency graph. The real clock retains UTC normalization and microsecond precision, with the corresponding assertion preserved.

## Validation

- `bazel test //services/tadoku-api/internal/timex:timex_unit_test //services/tadoku-api/e2e:e2e_test`
- `bazel test --@rules_go//go/config:race //services/tadoku-api/internal/timex:timex_unit_test`
- `bazel build //services/tadoku-api/cmd/tadoku-api:tadoku-api`
- production-to-test-helper `somepath` query returned empty results


services/tadoku-api/e2e/BUILD.bazel
```diff
@@ -49,6 +49,7 @@ go_test(
         "//services/tadoku-api/internal/testketo",
         "//services/tadoku-api/internal/testpostgres",
         "//services/tadoku-api/internal/timex",
+        "//services/tadoku-api/internal/timex/timextest",
```
antonve: this is wrong, it should just be timex_test package name inside the same package??

antonve:
I dont like it, closing

PR #905, opened 2026-09-17
fix(tadoku-api): map cancellation and deadline to their own statuses

antonve:
## Summary

- add an explicit immutable `errx.Internal` kind while leaving unclassified errors as `Unknown`
- map wrapped deadlines to 504 and canceled incoming requests to 499 before application-kind mapping
- use the incoming request context to distinguish request cancellation from a canceled dependency or child context
- keep ordinary PostgreSQL begin, commit, query, and decoding failures cause-preserving and unclassified, so they fall back to 500
- retain explicit 404 not-found, 409 conflict, and provider-boundary 503 behavior

Status 499 means cancellation was observed on the incoming request context. It does not claim perfect physical-client provenance because forced server connection closure also cancels request contexts. A cancellation from a dependency or child context while the request remains live follows its error kind: an unclassified error returns 500, while an explicitly unavailable provider error returns 503. Deadlines continue to return 504.

The operation logger and response handler use the same request-aware classifier. Canceled incoming requests are debug-level rejections; internal dependency cancellation, unavailable providers, deadlines, and other server failures are error-level failures. The golden update workflow remains explicitly opt-in and checks the expected status before writing.

Generic repository and transaction errors can represent constraint violations, query or decoding bugs, cancellation, connection failures, or an ambiguous commit outcome. These errors preserve their cause and use the conservative 500 fallback unless a more specific mapping applies. A cause-based PostgreSQL taxonomy is tracked separately in #915.

## Validation

- Regression before the cancellation fix: focused classifier tests compiled and failed because live-request raw and unavailable-wrapped cancellations both returned 499 instead of 500 and 503.
- Regression before the database-scope fix: focused tests compiled and failed because a deferred constraint failure at commit and a representative closed-pool repository read were both classified as `Unavailable` instead of `Unknown`.
- Passed focused request-status and log-severity tests, including 499, 500, 503, and 504 cases.
- Passed the focused repository and transaction error-kind regressions.
- Passed the affected dependency-failure and update-announcement E2Es with 500 goldens; the canceled-read E2E remains 499.
- Passed Gazelle diff and `git diff --check`.

Final commit validation passed: `bazel build //services/...`, `bazel test //services/... --test_output=errors` (all 39 service test targets), `bazel run //:gazelle -- -mode=diff`, and `bazel run //tools/ci/depolicy`.


services/tadoku-api/features/content/announcements_repository_test.go
```diff
@@ -10,10 +10,34 @@ import (
 	"github.com/google/uuid"
 	"github.com/tadoku/tadoku/services/tadoku-api/features/content"
 	"github.com/tadoku/tadoku/services/tadoku-api/infra/postgres"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/errx"
 	"github.com/tadoku/tadoku/services/tadoku-api/internal/testpostgres"
 	"github.com/tadoku/tadoku/services/tadoku-api/internal/timex"
 )
 
+func TestAnnouncementsRepositoryGenericDriverErrorIsUnclassified(t *testing.T) {
```
antonve, line 18: we dont need this

PR #911, opened 2026-09-17
fix(tadoku-api): bound readiness and pool acquisition

antonve:
## Problem

Database work could wait for the full request timeout when the PostgreSQL pool was exhausted, and readiness used that same request timeout. The initial fix exposed connection ownership to every repository method, requiring a `defer release()` beside each ordinary sqlc call.

## Change

Give pool acquisition a five-second budget while keeping SQL execution on the original request context. `postgres.Executor(ctx, pool)` keeps its normal `(DBTX, error)` shape and returns a sqlc-compatible pool executor outside transactions:

- `Exec` releases its connection when execution returns.
- `QueryRow` releases after `Scan`.
- `Query` releases when rows close or are exhausted.

Repositories use generated sqlc queries normally and do not manage connection ownership or wrap each operation in a callback. `RunInTransaction` still owns one acquired connection through begin, commit, rollback, and bounded cleanup.

Acquisition timeouts retain 503; unclassified database failures remain 500; existing domain mappings remain unchanged. Parent request cancellation and deadlines retain their existing 499 and 504 behavior.

Readiness has an independent two-second deadline and still checks a real PostgreSQL connection. The runtime pool keeps one minimum connection and explicitly uses pgx cached-statement mode.

## Validation

- `//services/tadoku-api/infra/postgres:postgres_test`
- `//services/tadoku-api/features/content:content_test`
- `//services/tadoku-api/transport/http:http_test`
- `//services/tadoku-api/e2e:e2e_test`
- `bazel test //services/... --test_output=errors` — all 39 targets pass
- Gazelle diff

Coverage includes connection release after `Exec`, `QueryRow.Scan`, early `Rows.Close`, and row exhaustion; SQL that runs longer than the acquisition budget; parent cancellation/deadline identity; transaction cleanup; pool saturation returning 503; and readiness returning promptly under saturation.

## Deployment gate

Production activation remains out of scope. Before deployment, verify the provider pooler version and prepared-statement configuration. Cached statements require prepared-statement support; if the deployed pooler cannot provide it, choose a different pgx query mode in a separate reviewed configuration change.


antonve:
Nvm this complexity, closing

PR #918, opened 2026-09-18
feat(tadoku-api): migrate blog post lookup

antonve:
Moves `GET /content/posts/{namespace}/{slug}` into Tadoku API. Published slugs remain public; the existing UUID fallback lets administrators retrieve drafts and scheduled posts. It preserves slug precedence, publication boundaries, namespace/deletion filtering, and authorization statuses.

The legacy response omits `created_at` and `updated_at` for slug lookup and includes them for admin ID fallback. Optional timestamp fields preserve that contract: the application clears timestamps after successful slug lookup and returns `(*content.Post, error)`, while transport maps the result directly. The content service retains the stored timestamps.

Layer 2 of the seven-endpoint blog migration stack. #917 has merged; this PR now targets `main`, and #919 builds on it.

Validation: `bazel build //services/...` and all 39 service test targets pass. Twenty-one shared HTTP cases compare native and legacy production handlers with real JWT, Keto, and PostgreSQL. Repository coverage checks current revision mapping, namespace isolation, nullable publication time, and transaction visibility. The lookup correction and downstream timestamp adaptations pass the existing goldens unchanged.


services/tadoku-api/app/post_find.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/google/uuid"
+	"github.com/tadoku/tadoku/services/tadoku-api/features/content"
+	"github.com/tadoku/tadoku/services/tadoku-api/internal/errx"
+)
+
+func (a *Application) FindPostBySlug(ctx context.Context, namespace, slug string) (*content.Post, bool, error) {
```
antonve: Not sure where or why you are doing this: but this is not following established patterns: DO NOT CHANGE THE RETURN TYPE, even more don't put that kind of logic in the transport layer???? why do you think we have the app and service layer?

antonve: I also need your motivation for changing this + why this behaves differently from the legacy API.

antonve: I added the `byID` boolean to preserve a legacy response difference: successful slug lookup omits `created_at` and `updated_at`, while successful admin ID fallback includes them. That was an unnecessary change to the return signature, and using the flag in transport put the field-selection policy in the wrong layer.

Fixed in b7b2643630ec558bedfda59817f45ab2d6c37b35: the app now returns `(*content.Post, error)`, the content service omits the optional timestamp fields for slug lookup, and transport maps the fields directly. ID fallback retains the timestamps. The dependent PRs have been restacked with the timestamp pointer adaptations.

The existing 21 lookup cases pass against both native and legacy production handlers with real JWT, Keto, and PostgreSQL, with unchanged HTTP goldens. The full service build and all 39 service test targets also pass locally.

**Posted by GPT-6**

antonve: Moved the `CreatedAt` / `UpdatedAt` clearing into the successful slug lookup branch of `Application.FindPostBySlug` in 01992380. The content service now returns the stored timestamps; admin ID fallback retains them, and transport remains direct field mapping.

Also added a deferred item in `services/tadoku-api/MIGRATION_LOG.md` to split all announcement, post, and page routes into separate admin and frontend routes after the migration is complete. #919–923 are restacked on these changes.

All 21 lookup parity cases pass with unchanged goldens. The full service build and all 39 service test targets pass locally.

**Posted by GPT-6**

PR #919, opened 2026-09-18
feat(tadoku-api): migrate blog post listing

antonve:
Moves `GET /content/posts/{namespace}` into Tadoku API. Public lists now hide scheduled posts until their publication time, including from totals; administrators using `include_drafts=true` can still see drafts and scheduled posts.

Layer 3 of the seven-endpoint blog migration stack. #918 has merged; this PR now targets `main`, and #920 builds on it. Approved behavior changes follow announcements: negative pagination returns 400, offsets cannot overflow, and equal creation timestamps sort by ID. Default/capped page sizes, empty results, current-content mapping, and the legacy guest/non-admin 403 for draft access are preserved.

Validation: `bazel build //services/...` and all 39 service test targets pass. Twenty-one HTTP cases use real JWT/Keto/PostgreSQL and shared legacy goldens, with explicit parity exceptions for the approved corrections. Repository tests verify publication boundaries, namespace/deletion filtering, stable paging, nullable fields, and totals on empty pages. SQL/OpenAPI/Gazelle outputs regenerated.


services/tadoku-api/app/post_list.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func (a *Application) ListPosts(ctx context.Context, namespace string, includeDrafts bool, pageSize, page int) (*content.PostList, error) {
+	if includeDrafts {
+		allowed, err := a.permissions.IsAdmin(ctx)
+		if err != nil {
+			return nil, err
+		}
+		if !allowed {
+			return nil, errx.NewForbiddenError("forbidden")
```
antonve: should be more specific "drafts require admin access" or something

antonve: Changed the error message to `drafts require admin access` in e293beb0. The existing ListPosts native/legacy HTTP cases pass, including guest and non-admin draft access returning 403, with unchanged goldens.

**Posted by GPT-6**

PR #940, opened 2026-09-19
Restructure Tadoku API features and validate announcement update IDs

antonve:
Feature operations were spread across individual files, and announcement tests mixed repository and service concerns. Group announcements, pages, and posts into five files each: service, service tests, repository, repository tests, and `domain.go` for shared types, errors, and validation.

Move all feature database tests into the repository test files and keep service/validation tests database-free. Preserve existing feature operations and coverage, update Bazel sources, and document the layout in AGENTS.md and the Tadoku API README. Keep all feature domain errors together in one declaration block per `domain.go`, including the former inline slug-validation and announcement-conflict errors. Their messages and error categories are unchanged.

Announcement updates now reject the zero UUID during parameter validation. A valid admin request to update `/content/announcements/main/00000000-0000-0000-0000-000000000000` returns HTTP 400 instead of the previous 404 lookup result. This intentional contract change has a database-free validation regression and a signed HTTP golden case; that case explicitly skips legacy parity because the legacy handler retains the 404 behavior.

Validation:

- `bazel build //services/...` passed on the final changes.
- All three feature test suites passed. The complete HTTP E2E suite passed when run separately, including parity and user journeys; its first concurrent run hit one request timeout (HTTP 504) during announcement creation.
- The zero-ID regression compiled and failed with `error=<nil>` before the validation fix, then passed afterward.
- The initial structural refactor passed `bazel run //:gazelle -- -mode=diff` and all 43 backend test targets after rebasing onto main; declarations and test bodies were compared against main before the validation change.


services/tadoku-api/features/announcements/domain.go
```diff
@@ -0,0 +1,110 @@
...
+		!strings.HasPrefix(*href, "//") && !strings.HasPrefix(*href, `/\`)
+}
+
+var (
+	ErrInvalidNamespace     = errx.NewInvalidInputError("namespace is required")
+	ErrInvalidPagination    = errx.NewInvalidInputError("invalid pagination")
+	ErrAnnouncementNotFound = errx.NewNotFoundError("announcement not found")
+)
+
+var ErrInvalidAnnouncement = errx.NewInvalidInputError("invalid announcement")
```
antonve: why is this the odd one out?

services/tadoku-api/features/announcements/domain.go
```diff
@@ -0,0 +1,110 @@
...
+	Namespace string
+	Title     string
+	Content   string
+	Style     string
+	Href      *string
+	StartsAt  time.Time
+	EndsAt    time.Time
+}
+
+func (p UpdateAnnouncementParameters) Validate() error {
```
antonve, line 103: we should actually validate that id is not a zero value

services/tadoku-api/features/announcements/domain.go
antonve: Grouped the announcement error in 2143cb28 and checked all three feature packages. Completed the remaining grouping in dbdddf03: all 17 feature domain errors now live in one `var` block per `domain.go`, including the former inline slug-validation and announcement-conflict errors. Messages and error categories are unchanged. The backend build, all feature suites, and the HTTP E2E suite pass.

**Posted by GPT-6**

services/tadoku-api/features/announcements/domain.go
antonve: Fixed in 2143cb28: `UpdateAnnouncementParameters.Validate()` now rejects `uuid.Nil`. The validation regression compiled and failed before the fix, then passed. A signed HTTP golden case verifies HTTP 400; it explicitly skips legacy parity because the legacy handler returns 404 for the zero ID.

**Posted by GPT-6**

PR #944, opened 2026-09-19
Wire raw Kratos SDK client into Tadoku API

antonve:
Tadoku API now constructs and retains a raw Kratos SDK client during normal startup, ready for explicit dependency injection. The shared `NewAPIClient` constructor reuses the pinned SDK while preserving existing `NewClient` callers and the SDK's models, responses, provider errors, and caller cancellation.

Both constructors accept `WithHTTPClient`. The shared wrapper's cursor pagination uses the SDK's configured HTTP client, so SDK and handwritten requests share the supplied transport and timeout without a duplicate HTTP-client field. Existing one-argument `NewClient` callers remain compatible.

`API_KRATOS_ADMIN_URL` is required and validated; `API_KRATOS_TIMEOUT` defaults to 2s and bounds the entire request, including response-body reads. The client uses the runtime's bounded HTTP transport, with cleanup on startup failure and shutdown. Development manifests use the existing Kratos admin service, and the README documents configuration and direct SDK usage. Startup makes no Kratos request. Domain mapping and consuming operations remain for future migrations; this change adds no application consumers, caches, refresh jobs, deployment, or live identity mutations.

Validation:

- `bazel build //services/...` — passed.
- `bazel test //services/... --test_output=errors` — all 43 test targets passed with disposable loopback PostgreSQL 17 and Valkey 9 instances.
- `bazel run //tools/ci/depolicy` — 136 Go files checked, zero violations.
- `bazel run //:gazelle -- -mode=diff` — clean after BUILD regeneration.
- `git diff --check` — clean.

Scoped HTTP fixtures cover SDK request options, custom TLS transport, provider models/status/headers/error bodies, caller cancellation, and client timeouts during headers and body reads. Shared-wrapper fixtures verify that both SDK and cursor-pagination requests use the supplied HTTP client and that pagination respects its timeout. Runtime tests cover configuration validation, request-free startup, direct use of the owned client, configured request bounds, failed startup without provider calls, and idle-connection cleanup on shutdown.


services/tadoku-api/cmd/tadoku-api/main.go
```diff
@@ -71,6 +78,20 @@ func loadConfig() (config, error) {
 	if err != nil || ketoURL.Host == "" || (ketoURL.Scheme != "http" && ketoURL.Scheme != "https") {
 		return config{}, fmt.Errorf("validate config: KetoReadURL must be an HTTP(S) URL")
 	}
+	kratosURL, err := url.Parse(cfg.KratosAdminURL)
+	if err != nil || kratosURL.Hostname() == "" || (kratosURL.Scheme != "http" && kratosURL.Scheme != "https") ||
+		kratosURL.User != nil || kratosURL.RawQuery != "" || kratosURL.ForceQuery || strings.Contains(cfg.KratosAdminURL, "#") {
+		return config{}, fmt.Errorf("validate config: KratosAdminURL must be an HTTP(S) URL without credentials, query or fragment")
+	}
+	if port := kratosURL.Port(); port != "" {
```
antonve: we are overly defensive here, we don't need to validate the port. we can just have the validation on the kratosURL

antonve: Removed the separate port-range and empty-port checks in 1e04d485, along with their test assertions and the README's port-validation claim. Kratos URL parsing and the existing HTTP(S), host, credentials, query and fragment checks remain.

`TestLoadConfigUsesValidatedDefaults` passes; Gazelle reports no diff and Depolicy reports zero violations.

**Posted by GPT-6**

PR #946, opened 2026-09-19
Seed RAM-backed Kratos once for Tadoku API E2Es

antonve:
HTTP E2Es now own a real Kratos v26.2.0 process with the standard four identities, seeded once in `TestMain`. Direct SQL seeding into private RAM-backed SQLite preserves the existing UUIDs, traits and timestamps, so signed requests, Keto relationships and golden responses remain unchanged.

Tests that mutate Kratos opt into cleanup: each native/legacy implementation sets `resetKratos: true`; a journey calls `resetKratosAfter(t, api)` on its enclosing test. Cleanup restores the full seed after that implementation or whole journey, including `t.Fatal` and cancelled test contexts. Ordinary PostgreSQL/Keto resets leave Kratos alone. Failed provider resets prevent subsequent scenarios from using the fixture.

The test-only helper pins the official SQLite-enabled binary and checksum, owns migrations, private Unix sockets, a bounded HTTP client passed to the existing raw SDK constructor, process reaping and file cleanup. Existing SDK clients survive an explicit reset. The identity schema is extracted from the dev Helm values and shared by Tilt and the fixture without changing its contents. README and AGENTS document usage and ownership.

This uses SQLite on `/dev/shm` tmpfs, rather than process-private `dsn: memory`, to allow fixed-ID seeding. It requires Linux x86-64 and writable tmpfs. Seed SQL is coupled to the pinned provider schema and creates identities only; credential/session scenarios must arrange their own provider state. This change adds test infrastructure, with no application consumers or live identity changes.

Approved design: https://draft.apps.lab/documents/tadoku-kratos-memory-e2e.html?version=3

Validation:

- `bazel build //services/...` — passed, 228 targets, repeated on the latest `main`.
- `bazel test //services/... --test_output=errors` — all 44 targets passed with disposable PostgreSQL 17 and Valkey 9; repeated on the final rebased branch (42 cached, 2 executed).
- `bazel test //services/common/client/kratos/... //services/tadoku-api/... --test_output=errors` — all 17 affected targets passed.
- `bazel test //services/tadoku-api/e2e:e2e_test --nocache_test_results --test_output=errors` — passed, including the subprocess proof that failed marked tests restore provider state.
- `bazel test //services/tadoku-api/internal/testkratos:testkratos_test --test_output=errors` — passed: fixed IDs, reset restoration, client reuse, fixture isolation, caller cancellation, partial startup cleanup and unexpected process exit.
- `bazel run --lockfile_mode=error //tools/ci/depolicy` — 140 Go files / 59 test files checked, zero violations on the final branch.
- `bazel mod deps --lockfile_mode=error`, `bazel run //:gazelle -- -mode=diff`, `./scripts/generate-sqlc.sh`, and `git diff --check` — passed; no stale generated output.
- Rendered the dev values with `k8s/dev/render_template.py`, then `helm template kratos ory/kratos --version 0.64.0` with the extracted schema supplied through `--set-file`; parsed JSON matches the original embedded schema exactly.
- Existing HTTP golden/token fixtures are unchanged; no owned Kratos tmpfs directories remain after test cleanup.


services/tadoku-api/e2e/testdata/kratos.sql
```diff
@@ -0,0 +1,15 @@
+-- Standard cast, matching the fixed subjects and traits in the signed JWTs.
+-- Kratos initializes the only network before this seed runs. Its admin API
+-- cannot choose identity UUIDs, so these test-only rows use its pinned schema.
+with actors(id, email, display_name) as (
+    values
+        ('11111111-1111-4111-8111-111111111111', 'reader@example.test', 'Reader One'),
```
antonve, line 6: this should be `User One`, update all tests and goldens to reflect this.

antonve: Deferred per your follow-up. Added the rename to [MIGRATION_LOG.md](https://github.com/tadoku/tadoku/blob/6435bef6a8567ea154751fe35f65e31b82183575/services/tadoku-api/MIGRATION_LOG.md#L14), covering Kratos seeds, tests, HTTP goldens, regenerated signed JWT fixtures and their public JWKS. Rebased after #952 merged; verification is left to CI.

**Posted by GPT-6**

PR #951, opened 2026-09-19
Migrate current role and public permission checks to Tadoku API

antonve:
Migrates `GET /authz/current-user/role` and `POST /authz/permission/check` into Tadoku API's authenticated strict handlers. Guest, user, and administrator role responses remain compatible; public permission checks retain the empty production allowlist, required-field validation, and denial behavior.

The shared native ban gate intentionally returns 403 where legacy current-role returns the banned role or legacy permission-check binding returns 400. Those differences are explicit `skipParity` cases. Empty-body permission requests preserve legacy authentication ordering through a narrowly scoped transport adapter.

Adds shared native/legacy HTTP goldens using production JWT and real Keto, plus the authorization visibility journey. Native runtime dependencies remain independent of legacy authz code.

Validation: full `bazel build //services/...` and `bazel test //services/...`; Tadoku API import policy; OpenAPI regeneration; Gazelle diff; reviewed golden regeneration and focused endpoint checks.

Stack base: `main`, including the merged Kratos fixture prerequisite (#946).


services/tadoku-api/e2e/authz_role_get_test.go
```diff
@@ -0,0 +1,33 @@
...
+		want        int
+		skipParity  string
+	}{
+		{description: []string{"guest"}, want: http.StatusOK},
+		{description: []string{"user"}, want: http.StatusOK},
+		{description: []string{"admin"}, want: http.StatusOK},
+		{
+			description: []string{"banned"},
+			want:        http.StatusForbidden,
+			skipParity:  "the native shared ban gate rejects banned callers while legacy returns their banned role",
```
antonve: this is wrong, it should return the banned role for banned users

services/tadoku-api/features/authz/authz_service.go
```diff
@@ -0,0 +1,48 @@
...
+func (s *Service) CheckPermission(ctx context.Context, parameters PermissionCheckParameters) error {
+	if err := s.permissions.RequireAuthenticated(ctx); err != nil {
+		return err
+	}
+	if err := parameters.Validate(); err != nil {
+		return err
+	}
+
+	// Production intentionally exposes no public permissions.
+	return errx.NewForbiddenError("forbidden")
```
antonve, line 50: don't we have an allow list of permissions one can check? why was that not ported over? (let's do this in a follow up PR in the stack though...)

services/tadoku-api/features/authz/domain.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+type PermissionCheckParameters struct {
+	Namespace string
+	Object    string
+	Relation  string
+}
+
+func (p PermissionCheckParameters) Validate() error {
+	if p.Namespace == "" || p.Object == "" || p.Relation == "" {
```
antonve: split these up so the error message is specific about which one is missing.

also update our agents documentation to prefer singular checks + error message rather than merging everything

services/tadoku-api/transport/http/authz_permission_check.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"io"
+	stdhttp "net/http"
+	"strings"
+
+	"github.com/tadoku/tadoku/services/tadoku-api/app"
+	"github.com/tadoku/tadoku/services/tadoku-api/generated/openapi"
+)
+
+func withAuthzPermissionCheckEmptyBodyCompatibility(next stdhttp.Handler) stdhttp.Handler {
```
antonve: we don't need to keep compatibility here.

services/tadoku-api/e2e/authz_role_get_test.go
antonve: Fixed in [91eb7b20](https://github.com/tadoku/tadoku/commit/91eb7b20). Current-role now returns 200 with `{"role":"banned"}` for banned users, including users who are also administrators. JWT authentication and the real ban lookup still run; only this exact GET may report the ban, and other protected routes still reject it. Native/legacy goldens and the role-visibility journey pass.

**Posted by gpt-6-astra**

services/tadoku-api/features/authz/domain.go
antonve: Fixed in [91eb7b20](https://github.com/tadoku/tadoku/commit/91eb7b20). Namespace, object, and relation now have separate validation branches and specific missing-field errors. AGENTS.md now prefers one condition per branch with an error identifying the field or rule. Validation tests pass.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/authz_permission_check.go
antonve: Removed the compatibility middleware in [91eb7b20](https://github.com/tadoku/tadoku/commit/91eb7b20). The generated decoder now returns 400 for an empty permission-check body, including a signed guest request. The native/legacy table explicitly records the intentional difference from legacy’s guest 401 response; JWT authentication still runs first. Endpoint and race checks pass.

**Posted by gpt-6-astra**

services/tadoku-api/features/authz/authz_service.go
antonve: The legacy allowlist machinery existed, but production passed an empty list; the initial port preserved only that deny-all behavior. Ported the machinery in the requested follow-up, [#960](https://github.com/tadoku/tadoku/pull/960), stacked on #957.

It matches configured namespace/relation pairs, asks Keto about the verified caller, and returns its actual allowed/denied result. Production still supplies an empty allowlist. Native/legacy goldens cover both decisions and independent namespace/relation restrictions; the full backend build and all 47 test targets pass.

**Posted by gpt-6-astra**

services/tadoku-api/features/authz/domain.go
antonve: Updated in [dd279944](https://github.com/tadoku/tadoku/commit/dd279944): each validation branch now calls the existing `errx.NewInvalidInputError` directly with its specific description. Removed the per-field sentinel errors and updated AGENTS.md to make that convention explicit. Existing validation and HTTP E2E tests pass.

**Posted by gpt-6-astra**

PR #953, opened 2026-09-19
Migrate contest discovery into Tadoku API

antonve:
Move contest listing, lookup, latest-official lookup, and configuration options from the immersion proxy into Tadoku API. The handlers use the existing production authentication and ban middleware, concrete feature repositories, and generated native SQL/HTTP bindings.

Preserve legacy filtering, pagination metadata, private/deleted visibility, organizer names, and activity/language ordering. Keep exact proxy routes for creation permissions and ongoing registrations so the new contest-ID route cannot intercept them before their migration.

Shared HTTP goldens run against both production handler stacks with real JWT verification, Keto, and PostgreSQL. Repository tests cover discovery queries and the shared cleanup preserves migration-seeded scoring rules.

Stack: targets `main`, which includes the merged Kratos fixture prerequisite #946. Next: #955.

Validation: `bazel build //services/...` and `bazel test //services/...` passed (46 test targets on the final integrated stack), including 24 shared contest HTTP cases. SQL/OpenAPI regeneration and Gazelle diff checks passed. Astra xhigh review findings are resolved. The full suite also verified unchanged content-endpoint parameter errors after the narrowly scoped immersion compatibility adjustment.


services/tadoku-api/internal/permissions/checker.go
```diff
@@ -83,6 +83,13 @@ func (c *Checker) IsAdmin(ctx context.Context) (bool, error) {
 	return allowed, nil
 }
 
+// IsAdminForPublicRead grants optional read visibility only after a successful
+// administrator lookup. Provider failures keep the request public.
+func (c *Checker) IsAdminForPublicRead(ctx context.Context) bool {
```
antonve: I don't like the suffix ForPublicRead, any better suggestions?

services/tadoku-api/transport/http/router.go
```diff
@@ -123,8 +125,15 @@ func NewHandler(
 		Middlewares: []openapi.MiddlewareFunc{
 			withJSONCharsetCompatibility,
 		},
-		ErrorHandlerFunc: func(w stdhttp.ResponseWriter, _ *stdhttp.Request, err error) {
-			writeJSON(w, stdhttp.StatusBadRequest, map[string]string{"message": err.Error()})
+		ErrorHandlerFunc: func(w stdhttp.ResponseWriter, request *stdhttp.Request, err error) {
+			message := err.Error()
+			var parameterErr *openapi.InvalidParamFormatError
+			if strings.HasPrefix(request.URL.Path, "/immersion/") &&
```
antonve: Why do we have this code here? This seems sus. And doesn't look like something I would want to keep in the Tadoku API long term in the native version.

services/tadoku-api/features/contests/contests_repository.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	languages, err := listContestLanguages(ctx, q, row.ID, row.LanguageCodeAllowList)
+	if err != nil {
+		return nil, err
+	}
+	return contestView(row, languages), nil
+}
+
+func (r *ContestsRepository) FindLatestOfficialContest(ctx context.Context) (*ContestView, error) {
```
antonve: We are starting to mix concerns in the repository layer and this is coming from the old Adoku Immersion API. The repository layer should be pretty dumb. It should just be fetching data from the database and then mapping it on top to our domain types. It shouldn't be composing any kind of queries. That should be done inside the service layer.

services/tadoku-api/features/contests/contests_repository.go
```diff
@@ -21,65 +21,55 @@ func NewContestsRepository(db *pgxpool.Pool) *ContestsRepository {
 	return &ContestsRepository{db: db}
 }
 
-func (r *ContestsRepository) ListContests(ctx context.Context, parameters ListParameters) ([]Contest, int, error) {
+func (r *ContestsRepository) CountContests(ctx context.Context, parameters ListParameters) (int, error) {
```
antonve: Are we doing the count of contests for pagination? If so, we should merge these into the original query because as mentioned in one of the PRs in the past, with read committed we can still have a different view of the data within the same transaction and have inconsistent counts.

services/tadoku-api/e2e/testdata/ListContests/400_malformed_page/golden.http
```diff
@@ -0,0 +1,6 @@
+>>> GET /immersion/contests?page=wat
+HTTP/1.1 400 Bad Request
+Content-Length: 131
+Content-Type: application/json; charset=UTF-8
+
+{"message":"Invalid format for parameter page: error binding string parameter: strconv.ParseInt: parsing \"wat\": invalid syntax"}
```
antonve: Should we be returning this specific error in a public response?

PR #955, opened 2026-09-19
Migrate contest creation permissions into Tadoku API

antonve:
Move `GET /immersion/contests/create-permissions` into Tadoku API. It reads account age from the owned Kratos SDK and checks the number of contests created in the current business year, preserving the legacy administrator bypass and exact response behavior.

The age boundary is inclusive and uses a calendar month. Young accounts return 500, exhausted quotas return 403, missing identities return 404, and signed guests retain the legacy JSON 500 response without a native panic. The canonical response documentation now describes the existing 404/500 behavior.

Eight shared HTTP cases compare the native and legacy production handlers using real JWT verification, Keto, PostgreSQL, and immutable seeded Kratos identities. The legacy Kratos constructor now forwards existing HTTP-client options so parity tests reuse its production mapping over the fixture transport. A repository test checks creation-year counting, including deleted contests.

Stack: depends on #953 and targets `t3code/contest-discovery`. Next: #958.

Validation: import-policy check, `bazel build //services/...`, and `bazel test //services/...` passed (46 test targets on the final integrated stack). SQL/OpenAPI regeneration and Gazelle diff checks passed. The documentation contract check, generated API check, typecheck, and production docs build also passed. Astra xhigh review approved the final slice and documentation correction.


services/tadoku-api/features/contests/contests_service.go
```diff
@@ -2,17 +2,54 @@ package contests
...
+	return &Service{
+		contests: repository,
+		kratos:   kratos,
+	}
 }
 
-func NewService(repository *ContestsRepository) *Service {
-	return &Service{contests: repository}
+func (s *Service) CheckCreatePermission(ctx context.Context, userID uuid.UUID) error {
+	identity, response, err := s.kratos.IdentityApi.GetIdentity(ctx, userID.String()).Execute()
```
antonve, line 29: Do we actually need to use Kratos for this? I feel like we could be using the identity from the context so if it's authenticated then we should be able to. this seems... ah I see so you want to get the created at time

PR #954, opened 2026-09-19
Migrate administrator user listing to Tadoku API

antonve:
Migrates administrator-only `GET /profile/users` into Tadoku API. It preserves pagination and fuzzy search, provider ordering, per-request role enrichment with administrator precedence, omitted default roles, and suppression of accepted account deletions.

The native cache reuses common Kratos cursor pagination, deduplicates identities, and refreshes every five minutes. Provider failures retain its previous snapshot; suppression-query failures clear it; learned suppressions remain sticky. Its background loop starts after successful construction and is canceled and joined before resources close.

Adds real-PostgreSQL suppression coverage, cache and domain checks, native/legacy HTTP goldens, and a user-listing journey. Parity uses the real legacy production cache through a narrow synchronous refresh entrypoint, with fresh cache state and the shared router. No schema migration or account-deletion worker is included.

Validation: backend build/tests, native/legacy E2Es, import policy, complete sqlc generation, OpenAPI generation, and Gazelle verification. Astra reviewed the source, fixtures, and generated changes.

Stack base: `t3code/migrate-authz-reads` (#951).

PR #957, opened 2026-09-20
Migrate administrator role updates to Tadoku API

antonve:
Moves `PUT /authz/users/{id}/role` into Tadoku API. Administrators can ban and unban existing users; authorization precedes domain validation, reasons retain the existing 1000-byte limit, and administrator targets remain protected. Keto is updated before the moderation audit is persisted, preserving the existing partial-failure behavior if the audit write fails.

The operation uses the native JWT middleware and shared ban gate. A banned caller's malformed request therefore returns 403 before body decoding; the legacy handler returns 400. The golden table marks this intentional difference with `skipParity`.

Native and legacy production handlers share full HTTP goldens with real JWT, Keto, PostgreSQL, and Kratos. The moderation journey lists users, bans a user, observes the updated listing and rejected access, unbans the user, and verifies restored access and audit records. A separate audit-failure journey proves that a failed audit write leaves the successful Keto ban in place. Repository coverage verifies persisted audit fields and metadata.

Validation: full backend build and tests, focused race tests, import-policy checks, and sqlc/OpenAPI/Gazelle generation checks.

Stacked on #954 (`t3code/migrate-profile-users`).


services/tadoku-api/app/authz.go
```diff
@@ -15,3 +17,9 @@ type PermissionCheckParameters = authz.PermissionCheckParameters
 func (a *Application) CheckPermission(ctx context.Context, parameters PermissionCheckParameters) error {
 	return a.authorization.CheckPermission(ctx, parameters)
 }
+
+type RoleUpdateParameters = authz.RoleUpdateParameters
+
+func (a *Application) UpdateRole(ctx context.Context, parameters RoleUpdateParameters) error {
```
antonve, line 36: CHECK PERMISSIONS IN APP LAYER ALWAYS UPDATE AGENTS.MD STOP PUTTING IT IN OTHER LAYERS

services/tadoku-api/features/authz/authz_repository.go
```diff
@@ -0,0 +1,47 @@
...
+
+type AuthzRepository struct {
+	db *pgxpool.Pool
+}
+
+func NewAuthzRepository(db *pgxpool.Pool) *AuthzRepository {
+	return &AuthzRepository{db: db}
+}
+
+func (r *AuthzRepository) CreateModerationAudit(ctx context.Context, audit ModerationAudit) error {
```
antonve: moderation audit should be part of a separate audit feature, not inside authz. Make sure to call all the features "Audit" instead of "ModerationAudit", we'll rename this table later. Add this rename to the migration todo log.

antonve: Actually, do we currently allow features to embed other features? I believe not, we might have to generate a new layer "components" which contain common functionality like audit. They're smaller scoped "features" that are meant to be embedded.

antonve: Decision from our discussion: compose two independent features in the application layer. `app.UpdateRole` will check administrator access, call the authz feature to change the role, and then call the audit feature to record the action. Audit owns its domain types and persistence; authz does not depend on audit. We will keep the existing feature boundaries without introducing a components layer.

Full endpoint access checks belong in app. Feature services may inspect permissions only to provide extended behavior within an already-authorized operation.

The existing Keto-before-audit order and partial-failure behavior stay intact. Native audit names will use Audit rather than ModerationAudit; renaming the existing database table will be tracked separately in MIGRATION_LOG.md.

**Posted by gpt-6-astra**

services/tadoku-api/app/authz.go
antonve: Fixed in [dec31ff6](https://github.com/tadoku/tadoku/commit/dec31ff6). Full operation access checks now live in app: administrator access for role updates, authentication for permission checks, and the missing-identity guard for current-role reads. Guest role reads remain allowed. Services retain permission facts and business rules, including target-admin protection. AGENTS.md and README now distinguish full endpoint access from permission checks that provide expanded behavior within an already-authorized operation. The existing authorization/validation-order goldens pass unchanged; all 48 backend test targets pass.

**Posted by gpt-6-astra**

services/tadoku-api/features/authz/authz_repository.go
antonve: Implemented the agreed two-feature composition in [dec31ff6](https://github.com/tadoku/tadoku/commit/dec31ff6). App checks access, calls authz to update Keto, then calls audit to record the event. The audit feature owns generic event types, timestamps, its repository, SQL, and complete generated SQLC output; authz has no audit dependency. Native types/query names use audit naming. The existing table remains `moderation_audit_log`, with its rename tracked in MIGRATION_LOG.md. Repository coverage and unchanged HTTP journeys verify stored fields, ordering, and audit failure without undoing Keto.

**Posted by gpt-6-astra**

services/tadoku-api/transport/http/authz_role_update.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"io"
+	stdhttp "net/http"
+	"strings"
+
+	"github.com/tadoku/tadoku/services/tadoku-api/app"
+	"github.com/tadoku/tadoku/services/tadoku-api/generated/openapi"
+)
+
+func withAuthzRoleUpdateEmptyBodyCompatibility(next stdhttp.Handler) stdhttp.Handler {
```
antonve: DO NOT WRITE MIDDLEWARE FOR EMPTY BODY COMPATIBILITY EVER, note that in agents.md

services/tadoku-api/app/authz.go
```diff
@@ -2,16 +2,67 @@ package app
...
+	if err := a.permissions.RequireAdmin(ctx); err != nil {
+		return err
+	}
+
+	actorID, err := uuid.Parse(identity.FromContext(ctx).Subject)
+	if err != nil {
+		return errx.NewUnauthorizedError("unauthorized")
+	}
+
+	if err := a.authorization.UpdateRole(ctx, parameters); err != nil {
```
antonve, line 46: this should be a transaction, but i believe this is an external system. can you add to migration log that we should be able to do a two-step audit log (start of action, and completion recorded)

services/tadoku-api/transport/http/authz_role_update.go
antonve: Removed the middleware and its registration in [55dd9234](https://github.com/tadoku/tadoku/commit/55dd9234). No replacement body rewrite was added. Empty-body role updates now use the generated decoder’s 400 response; the signed-guest difference from legacy 401 is explicit in `skipParity`. A valid-body guest case still verifies native/legacy 401 parity. AGENTS.md now prohibits request rewriting or decoder bypasses for empty-body compatibility. HTTP/E2E, race checks, and all 48 backend test targets pass.

**Posted by gpt-6-astra**

services/tadoku-api/app/authz.go
antonve: Added the migration TODO in [7955861f](https://github.com/tadoku/tadoku/commit/7955861f): persist an action’s start before calling the external system, then record its correlated completion and outcome. Keto cannot participate in the PostgreSQL transaction used for the audit write, so the current runtime behavior remains unchanged until that two-step design is implemented.

**Posted by gpt-6-astra**

PR #958, opened 2026-09-20
Migrate contest creation into Tadoku API

antonve:
Move `POST /immersion/contests` into Tadoku API, preserving legacy 200 responses, administrator rules, annual quotas, validation, user synchronization, and account-deletion conflicts. The user upsert still commits before quota/validation; the contest write and database readback run in an application-owned transaction with the user mutation lock.

Legacy creation now receives its UUID and timestamp through ordinary domain inputs using its existing clock. Native creation uses `timex`. This makes successful shared HTTP goldens deterministic without rewriting SQL or masking response fields. The canonical specification and generated docs also describe the existing malformed-identity JSON 500 response.

Eighteen shared HTTP cases run through both production handler stacks. Repository tests cover persistence and deletion-state behavior. The central creation journey verifies that a rejected create commits the user upsert without creating a contest, creates a member contest, reads it as another user, lists it, then creates an official contest and discovers it through latest-official.

Stack: depends on #955 and targets `t3code/contest-create-permissions`. Next: #959. No schema migration is included.

Validation: import-policy check, `bazel build //services/...`, and `bazel test //services/...` passed (46 test targets on the final integrated stack). SQL/OpenAPI regeneration and Gazelle diff checks passed; generated API documentation and its contract/type checks passed. Astra xhigh review findings are resolved.


services/tadoku-api/app/contests.go
```diff
@@ -29,6 +33,42 @@ func (a *Application) CheckContestCreatePermission(ctx context.Context) error {
 	return a.contests.CheckCreatePermission(ctx, userID)
 }
 
+func (a *Application) CreateContest(ctx context.Context, parameters CreateContestParameters) (*Contest, error) {
+	admin := false
+	if parameters.Official {
```
antonve: These nested if seemed very iffy. I think we can do better here.

services/tadoku-api/e2e/contest_create_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{description: []string{"banned"}, want: http.StatusForbidden},
+		{description: []string{"account", "deletion", "in", "progress"}, want: http.StatusConflict},
+	}
+
+	for _, test := range tests {
+		name := APITestName("CreateContest", test.want, test.description...)
+		t.Run(name, func(t *testing.T) {
+			runCase(t, api, name, test.want,
+				implementation{name: "tadoku-api", handler: api.handler, deterministicUUID: true},
+				implementation{name: "immersion-api", handler: legacyImmersion.handler, deterministicUUID: true},
```
antonve: Quick question: is there a case where we don't want a deterministic UUID?

services/tadoku-api/e2e/contest_create_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"net/http"
+	"testing"
+)
+
+func TestCreateContest(t *testing.T) {
+	tests := []struct {
+		description []string
+		want        int
+	}{
+		{description: []string{"member", "uses", "stored", "display", "and", "keeps", "duplicate", "allow", "lists"}, want: http.StatusOK},
```
antonve: Can you give me some more details on what this test actually tests? The title seems pretty cryptic for me.

services/tadoku-api/transport/http/contest_create.go
```diff
@@ -0,0 +1,63 @@
...
+		Description:             request.Body.Description,
+		Official:                request.Body.Official,
+		Private:                 request.Body.Private,
+		LanguageCodeAllowList:   request.Body.LanguageCodeAllowList,
+		ActivityTypeIDAllowList: request.Body.ActivityTypeIdAllowList,
+	}
+	contest, err := s.application.CreateContest(ctx, parameters)
+	if err != nil {
+		s.logOperationError(ctx, "create contest", err)
+		if errors.Is(err, app.ErrAccountDeletionInProgress) {
```
antonve: Shouldn't this be handled in our global error handler?

services/tadoku-api/sql/contests/contests.sql
```diff
@@ -76,6 +76,65 @@ from contests
 where owner_user_id = sqlc.arg(owner_user_id)
   and extract(year from created_at) = sqlc.arg(year)::integer;
 
+-- name: UpsertContestCreator :one
```
antonve: This should not be a context query. This should be in some kind of user profile kind of repository feature thing. And we should do the upsert in that service and it should be composed in the app list.

services/tadoku-api/features/contests/domain.go
```diff
@@ -88,6 +91,58 @@ type ContestList struct {
 	NextPageToken string
 }
 
+type CreateParameters struct {
```
antonve: create contest parameters.

services/tadoku-api/features/contests/domain.go
```diff
@@ -88,6 +91,58 @@ type ContestList struct {
...
+
+func (p CreateParameters) ID() uuid.UUID                { return p.id }
+func (p CreateParameters) OwnerUserID() uuid.UUID       { return p.ownerUserID }
+func (p CreateParameters) OwnerUserDisplayName() string { return p.ownerUserDisplayName }
+func (p CreateParameters) SessionCreatedAt() time.Time  { return p.sessionCreatedAt }
+func (p CreateParameters) CreatedAt() time.Time         { return p.createdAt }
+func (p CreateParameters) UpdatedAt() time.Time         { return p.updatedAt }
+
+func (p CreateParameters) validate(admin bool, now time.Time) error {
+	if p.ownerUserID == uuid.Nil || p.ownerUserDisplayName == "" ||
```
antonve: I believe we already have this documented in our agents.md file since this morning, but we should have error_invalid_contest: the specific thing that's invalid and it should be in the right golang syntax. However, this could potentially break parity with the old API. If that's the case, I want this item to be a migration to do for post migration.

services/tadoku-api/features/contests/domain.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"sort"
 	"time"
+	"unicode/utf8"
 
 	"github.com/google/uuid"
```
antonve, line 10: One thing I don't remember if we already have in the migration to do, but we should be using the standard UUID from the new Golang update 1.28, I guess. Don't change anything for it right now. I just want to make sure we don't forget that we have to make this change at some point in time.

services/tadoku-api/features/contests/contests_service.go
```diff
@@ -25,6 +26,62 @@ func NewService(repository *ContestsRepository, kratos *kratosapi.APIClient) *Se
 	}
 }
 
+func (s *Service) PrepareContestCreation(ctx context.Context, parameters CreateParameters, admin bool) (CreateParameters, error) {
+	creator := identity.FromContext(ctx)
```
antonve: The code in here seems a bit iffy, is there a way to clean this up more? Or with a break parity? Let's discuss.

services/tadoku-api/features/contests/contests_repository.go
```diff
@@ -37,6 +38,88 @@ func (r *ContestsRepository) CountContestsCreatedByUserForYear(ctx context.Conte
...
+	if errors.Is(err, pgx.ErrNoRows) {
+		return ErrAccountDeletionInProgress
+	}
+	if err != nil {
+		return fmt.Errorf("upsert contest creator: %w", err)
+	}
+	return nil
+}
+
+func (r *ContestsRepository) LockContestCreator(ctx context.Context, userID uuid.UUID) error {
```
antonve: Is this something that should be in the contest repository?

services/tadoku-api/features/contests/contests_repository.go
```diff
@@ -37,6 +38,88 @@ func (r *ContestsRepository) CountContestsCreatedByUserForYear(ctx context.Conte
 	return count, nil
 }
 
+func (r *ContestsRepository) UpsertContestCreator(ctx context.Context, parameters CreateParameters, now time.Time) error {
+	executor, err := postgres.Executor(ctx, r.db)
+	if err != nil {
+		return err
+	}
+	_, err = queries.New(executor).UpsertContestCreator(ctx, queries.UpsertContestCreatorParams{
```
antonve: Yeah, I think we already mentioned this somewhere else, but this should not be in ContestRepository.

services/tadoku-api/app/contests.go
```diff
@@ -5,28 +5,71 @@ import (
...
 
 func (a *Application) CheckContestCreatePermission(ctx context.Context) error {
 	if a.permissions.IsAdminOrFalse(ctx) {
 		return nil
 	}
 
-	user := identity.FromContext(ctx)
-	if user == nil {
-		return contests.ErrInvalidContestCreator
+	user, err := a.profile.SignedUser(ctx)
```
antonve: Can we record as a future migration to-do that we want to have permission to create a contest to be inside Keto? Now we have some kind of background job to keep that up to date.

services/tadoku-api/transport/http/errors.go
```diff
@@ -18,6 +20,19 @@ func (s *server) logOperationError(ctx context.Context, operation string, err er
 	s.logger.DebugContext(ctx, operation+" rejected", "error", err)
 }
 
+func writeResponseError(w stdhttp.ResponseWriter, request *stdhttp.Request, err error) {
+	switch {
+	case errors.Is(err, app.ErrAccountDeletionInProgress):
+		writeJSON(w, stdhttp.StatusConflict, openapi.ImmersionAccountDeletionInProgress{
+			Error: openapi.AccountDeletionInProgress,
+		})
+	case errors.Is(err, app.ErrInvalidContestCreator):
```
antonve: This one we can probably use like a special errx error and so we don't need to keep checking this.

services/tadoku-api/features/profile/profile_repository.go
```diff
@@ -0,0 +1,62 @@
...
+		return err
+	}
+	user, err := queries.New(executor).LockUser(ctx, pgtype.UUID{Bytes: userID, Valid: true})
+	if errors.Is(err, pgx.ErrNoRows) {
+		return ErrLocalUserNotFound
+	}
+	if err != nil {
+		return fmt.Errorf("lock local user: %w", err)
+	}
+	if user.DeletionLockedAt.Valid || user.DeletedAt.Valid {
```
antonve: I'm not sure if the location of this check is correct here or not.

services/tadoku-api/features/profile/profile_service.go
```diff
@@ -4,24 +4,53 @@ import (
...
 	return &Service{
-		cache: cache,
-		roles: roles,
+		repository: repository,
+		cache:      cache,
+		roles:      roles,
 	}
 }
 
+func (s *Service) SignedUser(ctx context.Context) (SignedUser, error) {
```
antonve: Is this really the right shape for this problem? To me it seems like identity from context already should solve what we're trying to solve here.

services/tadoku-api/features/contests/contests_service.go
```diff
@@ -25,6 +27,52 @@ func NewService(repository *ContestsRepository, kratos *kratosapi.APIClient) *Se
 	}
 }
 
+func (s *Service) PrepareContestCreation(
```
antonve: I don't really like the shape of this solution. It's combining validation and preparation of the request. And I don't like the shape.

services/tadoku-api/features/contests/contests_service.go
```diff
@@ -27,43 +27,50 @@ func NewService(repository *ContestsRepository, kratos *kratosapi.APIClient) *Se
...
 		}
 		if !exists {
-			return parameters, errx.NewInvalidInputError("invalid contest LanguageCodeAllowList: one or more languages do not exist")
+			return errx.NewInvalidInputError("invalid contest LanguageCodeAllowList: one or more languages do not exist")
 		}
 	}
+	return nil
+}
 
+func (s *Service) PrepareContestCreation(
```
antonve: there's no reason why this can't be done in the app layer this is just a stupid pass-by method now

